PR #7, opened 2017-02-22
Added Gaffer to list of related projects.

t616178:
Request to add Gaffer to list of related projects, as suggested by @keith-turner.


keith-turner:
Thanks.  I merged the change and updated the website.

PR #12, opened 2017-06-19
add jchevalay in contributors

josselinchevalay:


mjwall:
thanks.  would you mind sorting alphabetically by first name.  Also please verify your timezone, as is it would link to https://www.timeanddate.com/time/zones/fr, which doesn't appear valid.

josselinchevalay:
let me know if ok, sorry

mjwall:
Looks good, thanks.  We will get the merged in soon

PR #13, opened 2017-06-27
ACCUMULO-4668 Remove 1.6 from the downloads page

joshelser:


pages/downloads.md
```diff
@@ -116,23 +116,6 @@ For convenience, [MD5][MD5SUM_17] and [SHA1][SHA1SUM_17] hashes are also availab
 * [Examples][EXAMPLES_17]
 * [Javadoc][JAVADOC_17]
 
-### 1.6.6
```
mikewalch: If 1.6.6 is being removed from the downloads page, it should also be removed from the Releases dropdown in the navbar.  This can be done by editing `_includes/nav.html`

joshelser: Oops! Thanks for catching. Pushed a fix.

milleruntime:
Sweet good first step to retirement!

joshelser:
Thanks for taking a look, "Mikes" :)

PR #23, opened 2017-09-05
Update Jekyll gem versions

ctubbsii:
Updates json gem to 1.8.6 so it works on Ruby 2.4 and just generally bring gems up-to-date with latest github-pages and dependencies.


ctubbsii:
I did notice that the updated plugins generate anchor names slightly differently. This could break links within a page. I don't think this is a big problem, but something to look for.

ctubbsii:
Aww crap. There's some bug with `redirect_to`. It's encoding the "{{ site.baseurl }}/{{ site.latest_minor_release }}` directly into the redirect, rather than doing the substitution for `/1.8`. Perhaps @mikewalch can take a look, as I think he's a bit more familiar with the redirection plugins.

mikewalch:
@ctubbsii, I'll take a look this today

mikewalch:
It looks like variable substitution stopped working when the jekyll-redirect-from plugin was upgraded.  I tried several ways to get the substitution working again without luck so I created a PR to @ctubbsii branch that removes the variables.

PR #30, opened 2017-10-20
Updated 1.7 & 1.8 user manuals with latest changes

mikewalch:
Putting this up for PR in case anyone thinks there are changes that should wait for next release.


mikewalch:
I am going to close this PR.  The property `replication.rpc.timeout` was added.  This should not be in the manual until the next release.

PR #31, opened 2017-10-30
Update broken link on get_involved.md page

jmark99:
The link, "building open source communities", under the Community Interaction section is broken. I cannot access www.betaversion.org at all. I've updated the link with a document that might have been the original target of the link. If someone familiar with the page in the past can verify, we can use the updated link. Otherwise we should probably remove the "Community Interaction" section until we have content to place there.


jmark99:
I agree about the content. I would suggest we remove the Community Interaction sub-section until and if we get appropriate content for such a section. If there are no objections I will remove the sub-section altogether and then the page can be updated.

jmark99:
Community Interaction section has been removed and links adjusted. Should be good to go.

pages/get_involved.md
```diff
@@ -67,7 +63,7 @@ Most of the time we work with the consensus building techniques documented below
 
 ### Lazy Consensus
 
-[Lazy consensus][7] is the first, and possibly the most important, consensus building 
+[Lazy consensus][6] is the first, and possibly the most important, consensus building
```
ctubbsii: No need to renumber. The numbers don't matter. They are just identifiers, and it's okay to skip one. If you want to, instead of numbers, you can give them names, such as:

```
[Lazy consensus][lazy] is the first...
.
.
.
[lazy]: ...
```

jmark99: Thanks for the additional info. The linking format is new to me so additional info helps. Not sure about the additional copy of the file. I found the dead link as I was parsing through the various contributor info related pages. As I begin work on ACCUMULO-4714 I'll try to track and address issues like this. Thanks for the review.

get_involved.md
```diff
@@ -0,0 +1,94 @@
+---
+title: Get Involved
```
mikewalch: This PR adds a second `get_involved.md` page.  There is already one at at `pages/get_involved.md`.  This adds another at the root directory.  Pages were collected into `pages/` directory to reduce number of files in root directory.  The `permalink: /get_involved/` header will make sure page exists at root for URLs.

pages/get_involved.md
mikewalch: I will merge this in and switch from numbers to identifiers to follow @ctubbsii suggestion.  I will post the commit so you get an idea of what we are talking about

mikewalch: See d8338cc96c5a0eec

PR #35, opened 2017-11-03
ACCUMULO-4734 Content for getting-started and basic-read-write

milleruntime:
Looking for feedback before I start building further lessons on the first 2

### DIFF
--- a/tour/basic-read-write.md
+++ b/tour/basic-read-write.md
@@ -1,5 +1,67 @@
 ---
-title: Basic Reading & Writing
+title: Writing and Reading
 ---
+Accumulo is a big data key/value store.  Writing data to Accumulo is flexible and fast.  Like any database, Accumulo stores
+data in tables and rows.  Each row in an Accumulo table can hold many key/value pairs.  
 
-Talk about reading and writing.
+1. Start by connecting to Mini Accumulo and create a table called "superheroes".  For now, connect as the root user.
+```java
+Connector conn = mac.getConnector("root", "tourguide");
+conn.tableOperations().create("superheroes");
+```
+
+2. Create a Mutation object for row1
+```java
+Mutation mutation = new Mutation("row1");
+```
+A Mutation is an object that holds all changes to a row in a table.  Each row has a unique row ID.
+ 
+3. Create key/value pairs for Batman.
+```java
+mutation.put("name", "", "Batman");
+mutation.put("real-name", "", "Bruce Wayne");
+mutation.put("wearsCape?", "", "true");
+mutation.put("flies?", "", "false");
+```
+Every Mutation in Accumulo is atomic. This means that all the changes to a single row will happen at once. The Mutation
+object conveniently allows us to put all the changes for the row in one spot. 
+
+4. Create a BatchWriter to the superhero table and add your mutation to it.
+```java
+BatchWriter writer = conn.createBatchWriter("superheroes", new BatchWriterConfig());
+writer.addMutation(mutation);
+writer.close();
+```
+Accumulo is very efficient, so it prefers data to arrive in batches.  The BatchWriter will take care of this for us.
+Once we are finished, closing the BatchWriter will tell Accumulo everything is good to go.
+
+5. Print all rows of the "superheroes" table
+```java
+Scanner scan = conn.createScanner("superheroes", Authorizations.EMPTY);
+System.out.println("superheroes table contents:");
+for(java.util.Map.Entry entry : scan) {
+    System.out.println("Key:" + entry.getKey());
+    System.out.println("Value:" + entry.getValue());
+}
+```
+A Scanner is the object that Accumulo clients use to read data efficiently. A Scanner is an extension of 
+java.util.Iterator so behaves just like one. More on Scanners later.  
+
+
+6. Build and run your code
+```commandline
+mvn -q clean compile exec:java
+``` 
+
+Good job!  That is all it takes to write and read from Accumulo.  
+
+Notice a lot of other information was printed from the Keys we created. Accumulo is flexible because hidden within its 
+Key is a rich data model that can be broken up into different parts.  We will cover the [Data Model][dmodel] in the next lesson.
+
+### But wait... I thought Accumulo was all about Security?  
+Spoiler Alert: it is!  Did you notice the _Authorizations.EMPTY_ we passed to the Scanner on step 5?  The data
+we created in this first lesson was not secured with Authorizations so the Scanner didn't require any Authorizations 
+to read it.  More to come later in the [Authorizations][auths] lesson! 
+
+[dmodel]: /tour/data-model
+[auths]: /tour/authorizations
--- a/tour/basic-read-write.md
+++ b/tour/basic-read-write.md
@@ -4,51 +4,45 @@ title: Writing and Reading
 Accumulo is a big data key/value store.  Writing data to Accumulo is flexible and fast.  Like any database, Accumulo stores
 data in tables and rows.  Each row in an Accumulo table can hold many key/value pairs.  
 
-1. Start by connecting to Mini Accumulo and create a table called "superheroes".  For now, connect as the root user.
+Copy and paste the code below into the _exercise_  method.
 ```java
-Connector conn = mac.getConnector("root", "tourguide");
-conn.tableOperations().create("superheroes");
+        // 1. Start by connecting to Mini Accumulo as the root user and create a table called "superheroes".
+        Connector conn = mac.getConnector("root", "tourguide");
+        conn.tableOperations().create("superheroes");
+
+        // 2. Create a Mutation object to write to a row
+        Mutation mutation = new Mutation("hero023948092");
+        // A Mutation is an object that holds all changes to a row in a table.  Each row has a unique row ID.
+
+        // 3. Create key/value pairs for Batman.  Put them in the "HeroAttribute" family.
+        mutation.put("HeroAttribute","name", "Batman");
+        mutation.put("HeroAttribute","real-name", "Bruce Wayne");
+        mutation.put("HeroAttribute","wearsCape?", "true");
+        mutation.put("HeroAttribute","flies?","false");
+
+        // 4. Create a BatchWriter to the superhero table and add your mutation to it.  Try w/ resources will close for us.
+        try(BatchWriter writer = conn.createBatchWriter("superheroes", new BatchWriterConfig())) {
+            writer.addMutation(mutation);
+        } catch(TableNotFoundException | MutationsRejectedException e) {
+            System.out.println("Error in the BatchWriter:");
+            e.printStackTrace();
+        }
+
+        // 5. Read and print all rows of the "superheroes" table. Try w/ resources will close for us.
+        try(Scanner scan = conn.createScanner("superheroes", Authorizations.EMPTY)) {
+            System.out.println("superheroes table contents:");
+            // A Scanner is an extension of java.lang.Iterable so behaves just like one.
+            for (Map.Entry<Key, Value> entry : scan) {
+                System.out.println("Key:" + entry.getKey());
+                System.out.println("Value:" + entry.getValue());
+            }
+        } catch(TableNotFoundException e) {
+            System.out.println("Error performing scan:");
+            e.printStackTrace();
+        }
 ```
 
-2. Create a Mutation object for row1
-```java
-Mutation mutation = new Mutation("row1");
-```
-A Mutation is an object that holds all changes to a row in a table.  Each row has a unique row ID.
- 
-3. Create key/value pairs for Batman.
-```java
-mutation.put("name", "", "Batman");
-mutation.put("real-name", "", "Bruce Wayne");
-mutation.put("wearsCape?", "", "true");
-mutation.put("flies?", "", "false");
-```
-Every Mutation in Accumulo is atomic. This means that all the changes to a single row will happen at once. The Mutation
-object conveniently allows us to put all the changes for the row in one spot. 
-
-4. Create a BatchWriter to the superhero table and add your mutation to it.
-```java
-BatchWriter writer = conn.createBatchWriter("superheroes", new BatchWriterConfig());
-writer.addMutation(mutation);
-writer.close();
-```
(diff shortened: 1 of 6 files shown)


tour/basic-read-write.md
```diff
@@ -1,5 +1,67 @@
...
-title: Basic Reading & Writing
+title: Writing and Reading
 ---
+Accumulo is a big data key/value store.  Writing data to Accumulo is flexible and fast.  Like any database, Accumulo stores
+data in tables and rows.  Each row in an Accumulo table can hold many key/value pairs.  
 
-Talk about reading and writing.
+1. Start by connecting to Mini Accumulo and create a table called "superheroes".  For now, connect as the root user.
+```java
+Connector conn = mac.getConnector("root", "tourguide");
```
keith-turner: I would recommend restructuring this so that the code is all together and can be copied and pasted.  Anything explanation could be comments in the code.

tour/basic-read-write.md
```diff
@@ -1,5 +1,67 @@
...
+conn.tableOperations().create("superheroes");
+```
+
+2. Create a Mutation object for row1
+```java
+Mutation mutation = new Mutation("row1");
```
keith-turner: I would recommend a user id for the row like `hero342897`

tour/basic-read-write.md
```diff
@@ -1,5 +1,67 @@
...
+BatchWriter writer = conn.createBatchWriter("superheroes", new BatchWriterConfig());
+writer.addMutation(mutation);
+writer.close();
+```
+Accumulo is very efficient, so it prefers data to arrive in batches.  The BatchWriter will take care of this for us.
+Once we are finished, closing the BatchWriter will tell Accumulo everything is good to go.
+
+5. Print all rows of the "superheroes" table
+```java
+Scanner scan = conn.createScanner("superheroes", Authorizations.EMPTY);
```
keith-turner: I would use try with resources.  Good practice.  Like the following

```java
try(Scanner scanner = conn.createScanner("superheroes", Authorizations.EMPTY)){
  System.out.println("superheroes table contents:");
  for(Entry<Key, Value> entry : scan) {
￼     System.out.println("Key:" + entry.getKey());
￼     System.out.println("Value:" + entry.getValue());
  }
}
```

I would also use the type params for Entry and not upt the full package for Entry.

tour/basic-read-write.md
```diff
@@ -1,5 +1,67 @@
...
+mutation.put("real-name", "", "Bruce Wayne");
+mutation.put("wearsCape?", "", "true");
+mutation.put("flies?", "", "false");
+```
+Every Mutation in Accumulo is atomic. This means that all the changes to a single row will happen at once. The Mutation
+object conveniently allows us to put all the changes for the row in one spot. 
+
+4. Create a BatchWriter to the superhero table and add your mutation to it.
+```java
+BatchWriter writer = conn.createBatchWriter("superheroes", new BatchWriterConfig());
```
keith-turner: I would use try with resources with batch writer also.

tour/basic-read-write.md
```diff
@@ -1,5 +1,67 @@
...
+```java
+Scanner scan = conn.createScanner("superheroes", Authorizations.EMPTY);
+System.out.println("superheroes table contents:");
+for(java.util.Map.Entry entry : scan) {
+    System.out.println("Key:" + entry.getKey());
+    System.out.println("Value:" + entry.getValue());
+}
+```
+A Scanner is the object that Accumulo clients use to read data efficiently. A Scanner is an extension of 
+java.util.Iterator so behaves just like one. More on Scanners later.
```
keith-turner: I think a scanner is an Iterable not Iterator.

tour/basic-read-write.md
```diff
@@ -1,5 +1,67 @@
...
+```java
+mutation.put("name", "", "Batman");
+mutation.put("real-name", "", "Bruce Wayne");
+mutation.put("wearsCape?", "", "true");
+mutation.put("flies?", "", "false");
+```
+Every Mutation in Accumulo is atomic. This means that all the changes to a single row will happen at once. The Mutation
+object conveniently allows us to put all the changes for the row in one spot. 
+
+4. Create a BatchWriter to the superhero table and add your mutation to it.
```
keith-turner: Could say " Write the mutation to the superhero table."  The fact that a batch writer is used to do is self evident from the code.

tour/basic-read-write.md
```diff
@@ -1,5 +1,67 @@
...
+```java
+BatchWriter writer = conn.createBatchWriter("superheroes", new BatchWriterConfig());
+writer.addMutation(mutation);
+writer.close();
+```
+Accumulo is very efficient, so it prefers data to arrive in batches.  The BatchWriter will take care of this for us.
+Once we are finished, closing the BatchWriter will tell Accumulo everything is good to go.
```
keith-turner: Could mention that only after close is called, do you know the data is written.

keith-turner: could also cover this information later in the tour... keep the starting page simple

tour/basic-read-write.md
milleruntime: Yeah you are correct. Will fix

milleruntime:
Yep these are all great thanks @keith-turner

tour/basic-read-write.md
```diff
@@ -1,5 +1,61 @@
...
+        // 3. Create key/value pairs for Batman.  Put them in the "HeroAttribute" family.
+        mutation.put("HeroAttribute","name", "Batman");
+        mutation.put("HeroAttribute","real-name", "Bruce Wayne");
+        mutation.put("HeroAttribute","wearsCape?", "true");
+        mutation.put("HeroAttribute","flies?","false");
+
+        // 4. Create a BatchWriter to the superhero table and add your mutation to it.  Try w/ resources will close for us.
+        try(BatchWriter writer = conn.createBatchWriter("superheroes", new BatchWriterConfig())) {
+            writer.addMutation(mutation);
+        } catch(TableNotFoundException | MutationsRejectedException e) {
```
keith-turner: I think it would be nice to make the exercise method throw Exception and drop this catch block.  Doing this makes this first bit of code shorter someone sees shorter, which I think is important.

tour/basic-read-write.md
```diff
@@ -1,5 +1,61 @@
...
+
+        // 5. Read and print all rows of the "superheroes" table. Try w/ resources will close for us.
+        try(Scanner scan = conn.createScanner("superheroes", Authorizations.EMPTY)) {
+            System.out.println("superheroes table contents:");
+            // A Scanner is an extension of java.lang.Iterable so behaves just like one.
+            for (Map.Entry<Key, Value> entry : scan) {
+                System.out.println("Key:" + entry.getKey());
+                System.out.println("Value:" + entry.getValue());
+            }
+        } catch(TableNotFoundException e) {
```
keith-turner: If exercise throws exception then could also drop this catch block.

tour/basic-read-write.md
```diff
@@ -1,5 +1,61 @@
...
+        // 1. Start by connecting to Mini Accumulo as the root user and create a table called "superheroes".
+        Connector conn = mac.getConnector("root", "tourguide");
+        conn.tableOperations().create("superheroes");
+
+        // 2. Create a Mutation object to write to a row
+        Mutation mutation = new Mutation("hero023948092");
+        // A Mutation is an object that holds all changes to a row in a table.  Each row has a unique row ID.
+
+        // 3. Create key/value pairs for Batman.  Put them in the "HeroAttribute" family.
+        mutation.put("HeroAttribute","name", "Batman");
```
keith-turner: Was thinking something shorter for the col family would be nice and thought of `hero`. This made me think only things applicable to a hero would be under that family.  More general info that is applicable to anyone could be under a `person` family.  This would allow non-heros and villains to be in the DB also, but that makes me wonder what the DB should be called. This made me wonder who owns the DB?  I thought of Gotham PD, some PI, some hero, some villain, or a news org owning the DB.  I randomly chose GPD.

```java
 conn.tableOperations().create("gpd-poi");  //gotham PD persons of interest
  // 2. Create a Mutation object to write to a row
 Mutation mutation = new Mutation("id023948092");
mutation.put("hero","alias", "Batman");
￼mutation.put("hero","wearsCape?", "true");
mutation.put("hero","flies?","false");
mutation.put("person","name", "Bruce Wayne");
mutation.put("person","DOB", "XX-YY-ZZZZ");
```

Thinking ahead I like this example because it allows future visibility labels like the following.  The visibility labels will ultimately be tied to who owns the db.

```java
 conn.tableOperations().create("gpd-poi");  //gotham PD persons of interest
  // 2. Create a Mutation object to write to a row
 Mutation mutation = new Mutation("id023948092");
mutation.put("hero","alias", "secretId", "Batman");
￼mutation.put("hero","wearsCape?" "secretId", "true");
mutation.put("hero","flies?" "secretId & ability","false");
mutation.put("person","name", "Bruce Wayne");
mutation.put("person","DOB" ,"pii", "XX-YY-ZZZZ");
```

All of the above is me just thinking out loud, do what you will with it.

tour/basic-read-write.md
milleruntime: It does throw Exception, mainly because of the Connector.  For some reason I felt compelled to have a catch.  I guess it just feels wrong to have a try without a catch haha.  But it would greatly shorten the code.

tour/basic-read-write.md
milleruntime: This is what I was thinking for the lesson on Authorizations.  I wanted to keep the first one as simple as possible.

I was also thinking about keeping the the table in the first exercise called "superheroes" and making it more generic in follow on exercises.  The idea being a traditional table (in a relational DB) would be called superheroes with the DB being named "Gotham PD POI" and then explaining how Accumulo stores data differently aka a single table being more like a DB.   Something like.. you may think creating separate "villian" and "hero" tables would be a good idea but Accumulo allows you to store both in the same table and differentiate using Column Families.

keith-turner: I was not thinking of putting auths in this example.  I was just thinking ahead to when we use auths, it would be nice to be consistent and use the same data model.

tour/getting-started.md
```diff
@@ -2,4 +2,29 @@
...
 
-Talk about how to get started.
+First make sure you have Java, Maven and Git installed on your machine.  Oh you are already rocking? OK let's go!
+
+1. Clone the tour onto your machine:
+```commandline
+git clone -b tour https://github.com/apache/accumulo-website.git tour
+cd tour
+```
+2. Open Main.java in your favorite editor.
```
mikewalch: Could make Main.java link to https://github.com/apache/accumulo-website/blob/tour/src/main/java/tour/Main.java

tour/getting-started.md
```diff
@@ -2,4 +2,29 @@
...
+```commandline
+vim ./src/main/java/tour/Main.java
+```
+Notice the main method creates a MiniAccumuloCluster with a root password of "tourguide".  MiniAccumuloCluster is a mini
+version of Accumulo that runs on your local filesystem.  It should only be used for development purposes but will work
+great here on the tour.
+
+3. Modify the _exercise_ method to print a hello message. You will put your code in this method for each lesson.
+```java
+private static void exercise(MiniAccumuloCluster mac) {
```
mikewalch: could remove `private`

tour/getting-started.md
```diff
@@ -2,4 +2,29 @@
...
+1. Clone the tour onto your machine:
+```commandline
+git clone -b tour https://github.com/apache/accumulo-website.git tour
+cd tour
+```
+2. Open Main.java in your favorite editor.
+```commandline
+vim ./src/main/java/tour/Main.java
+```
+Notice the main method creates a MiniAccumuloCluster with a root password of "tourguide".  MiniAccumuloCluster is a mini
```
mikewalch: Could mention  that it stores files in `target/` in temp that starts with `mac`.  Could also add comment about this in tour code.

tour/getting-started.md
```diff
@@ -2,4 +2,29 @@
...
+great here on the tour.
+
+3. Modify the _exercise_ method to print a hello message. You will put your code in this method for each lesson.
+```java
+private static void exercise(MiniAccumuloCluster mac) {
+    // start writing your code here
+    System.out.println("Hello world");
+}
+```
+4. Build and run to make sure everything is cool.
```
mikewalch: could be `make sure everything works.`

tour/getting-started.md
```diff
@@ -2,4 +2,29 @@
...
+```
+Notice the main method creates a MiniAccumuloCluster with a root password of "tourguide".  MiniAccumuloCluster is a mini
+version of Accumulo that runs on your local filesystem.  It should only be used for development purposes but will work
+great here on the tour.
+
+3. Modify the _exercise_ method to print a hello message. You will put your code in this method for each lesson.
+```java
+private static void exercise(MiniAccumuloCluster mac) {
+    // start writing your code here
+    System.out.println("Hello world");
```
mikewalch: Could use logger here instead

keith-turner: would need to ensure `mvn -q` does not suppress logging output.

tour/getting-started.md
milleruntime: Yeah good call.  I put this in the README but wouldn't hurt to mention on the tour as well.

tour/getting-started.md
milleruntime: I wasn't going to mention logging since its not specific to Accumulo or needed here.  

The mvn -q won't suppress logging.  You can actually see the INFO log about MAC starting up.

keith-turner: > I wasn't going to mention logging since its not specific to Accumulo or needed here.

I agree with this this.. nice to keep it simple and focused on Accumulo APIs.

mikewalch: A logger doesn't need to be used but I suggested it because Accumulo still outputs log info messages.  I think it's confusing and hard to read when log messages and Sys.out.println is mixed (unless you are using Sys.out.prinln to output data)

milleruntime:
Thanks for the feedback.  I am going to close this for now until I have more of the exercises done.

PR #38, opened 2017-11-09
Continue tour

milleruntime:
Exercises added for:
Auths
data model
ranges and splits
batch scanner

### DIFF
--- a/tour/authorizations.md
+++ b/tour/authorizations.md
@@ -1,9 +1,42 @@
 ---
 title: Authorizations
 ---
+An Authorization is an authority granted to a User that allows them to read certain data. A column in Accumulo can have 
+a visibility attached to it, limiting access to the data of that column. Only users who have an authorization equal to 
+the visibility will be able to read the data in that column.
 
+We now want to secure our secret identities of the heroes so that only users with the proper authorizations can read their names.
 
-X. Build and run to make sure everything is peachy.
+1. Using the code from the previous exercise, add the following to the beginning of the _exercise_ method (after we get the Connector).
+```java
+        // Create a "secretIdentity" authorization & visibility
+        final String secId = "secretIdentity";
+        Authorizations auths = new Authorizations(secId);
+        ColumnVisibility visibility = new ColumnVisibility(secId);
+        
+        // Create a user with the "secretIdentity" authorization and grant him read permissions on our table
+        conn.securityOperations().createLocalUser("commissioner", new PasswordToken("gordanrocks"));
+        conn.securityOperations().changeUserAuthorizations("commissioner", auths);
+        conn.securityOperations().grantTablePermission("commissioner", "GothamPD", TablePermission.READ);
+``` 
+
+2. The Mutation API allows you to set the visibility on a column. Find the proper method for setting a column visibility in 
+the [Mutation API][mut] and modify the code so the visibility created above will secure the two "name" columns. 
+
+3. Build and run.  What data do you see?
+* You should see all of the data except the secret identities of Batman and Robin.  This is because the Scanner was created
+ from the root user Connector.  
+* Replace the _Authorizations.EMPTY_ in the Scanner with the _auths_ created above and run it again.
+* This should result in an error since the root user doesn't have the authorizations we tried to pass to the Scanner.
+
+4. Get a connector for the "commissioner" and from it create a Scanner with the authorizations needed to view the secret identities.
+
+5. Build and run.  You should see all the rows in the GothamPD table printed, including these secured key/value pairs:
 ```commandline
-mvn -q clean compile exec:java
+Key:id0001 hero:name [secretIdentity] 1510092746436 false
+Value:Bruce Wayne
+Key:id0002 hero:name [secretIdentity] 1510092746436 false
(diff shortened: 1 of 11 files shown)


tour/ranges-splits.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+As your data grows larger, Accumulo will split tables across multiple servers called Tablet Servers.  
+By default a table will get split on row boundaries, guaranteeing an entire row to be on one Tablet Server.  We have the ability to 
+tell Accumulo were to split tables by setting split points. This is done using _addSplits_ in the [TableOperations] API.  The image below 
+demonstrates how Accumulo splits data.  
+
+![data distribution]({{ site.url }}/images/docs/data_distribution.png)
+
+There are many useful administrative methods in [TableOperations] so take a minute to look through the API. Here are some terms specific to Accumulo:
```
keith-turner: This is good content, but it feels out of place here.

tour/ranges-splits.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Ranges and Splits
+---
+
+A Range is a specified group of Keys. There are many different ways to create a Range.  Here are a few examples:
+```java
+new Range(Key startKey, Key endKey)  // Creates a range from startKey inclusive to endKey inclusive.
+new Range(CharSequence row)  // Creates a range that covers an entire row.
```
keith-turner: Its odd that Jekyll renders this `Range` as red but not the one on the line before.

tour/authorizations.md
```diff
@@ -0,0 +1,42 @@
+---
+title: Authorizations
+---
+An Authorization is an authority granted to a User that allows them to read certain data. A column in Accumulo can have 
+a visibility attached to it, limiting access to the data of that column. Only users who have an authorization equal to 
+the visibility will be able to read the data in that column.
```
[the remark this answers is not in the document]
milleruntime: That does sound better but I was trying to make it clear that authorization = visibility.  This was something that tripped me up as a beginner, they have different names for different reasons but it isn't clear anywhere that it is a simple string compare.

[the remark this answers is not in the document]
keith-turner: One is a set and one is an expression.  Could say a column visibility is a boolean expression that is evaluated using a scanners authorizations.  If it evaluates to true, then the data is visible.

tour/authorizations.md
```diff
@@ -1,9 +1,15 @@
...
-An Authorization is an authority granted to a User that allows them to read certain data. A column in Accumulo can have 
-a visibility attached to it, limiting access to the data of that column. Only users who have an authorization equal to 
-the visibility will be able to read the data in that column.
+Authorizations are a set of Strings that enable a user to read protected data. A column visibility is a boolean expression 
+that is evaluated using the authorizations provided by a scanner. If it evaluates to true, then the data is visible. 
+
+For example:
+* Bob has authorizations = { IT, User }
+* Tina has authorizations = { Admin, IT, User }
+* Row1:A:B has Visibility = { Admin && IT && User }
```
keith-turner: Could change `A:B` to `family1:qualifier1`.   This makes the line longer, but if may make it more clear what the fields are.

tour/authorizations.md
```diff
@@ -1,9 +1,15 @@
...
 title: Authorizations
 ---
-An Authorization is an authority granted to a User that allows them to read certain data. A column in Accumulo can have 
-a visibility attached to it, limiting access to the data of that column. Only users who have an authorization equal to 
-the visibility will be able to read the data in that column.
+Authorizations are a set of Strings that enable a user to read protected data. A column visibility is a boolean expression 
+that is evaluated using the authorizations provided by a scanner. If it evaluates to true, then the data is visible. 
+
+For example:
+* Bob has authorizations = { IT, User }
```
keith-turner: This example is nice.   Could say `Bob scans with authorizations`

tour/data-model-code.md
```diff
@@ -35,31 +38,21 @@ title: Data Model Code
...
-Value:Dick Grayson
-Key:id0002 hero:wearsCape? [] 1510084285325 false
-Value:true
-Key:id0003 villain:alias [] 1510084285325 false
-Value:Joker
-Key:id0003 villain:name [] 1510084285325 false
-Value:Unknown
-Key:id0003 villain:wearsCape? [] 1510084285325 false
-Value:false
+Key : id0001 hero:alias [] 1511306370025 false            Value : Batman
```
keith-turner: The timestamp and delete boolean makes the output noisy.  However trying to omit would make the code noisy/longer and harder to understand.  I can't think of any way to improve this though.

tour/ranges-splits.md
milleruntime: I agree. I was trying to point out the more useful API classes and the best places to start.  Maybe we could do that in another exercise

tour/ranges-splits.md
milleruntime: I was going to merge the content I had and then let @mikewalch clean it up.  Maybe he can figure out why.

milleruntime:
I am merging this up to the branch so others can continue work on the tour.

PR #41, opened 2017-11-27
ACCUMULO-4734 Updated syntax highlighting

mikewalch:


mikewalch:
@ctubbsii, this is the color scheme that Apache Fluo uses.  @milleruntime, made a comment that it's more readable and I agree. Below is an example of the Fluo syntax highlighting:

https://fluo.apache.org/docs/fluo/1.1.0-incubating/applications/ 

Below is an example of current highlighting:

https://accumulo.apache.org/docs/2.0/development/mapreduce

PR #42, opened 2017-11-29
ACCUMULO-WEBSITE Correct typo on range-splits page

jmark99:
ACCUMULO-WEBSITE Correct typo on range-splits page.

Found a small typo on range-splits.md page when going through Accumulo Tour examples.
Corrected  from  'were' to 'where'


milleruntime:
Thanks Mark!

PR #45, opened 2017-12-04
ACCUMULO-4747 Create a unified upgrade reference

jmark99:
ACCUMULO-4747 Create a unified upgrade reference

Based on discussion with several other developers, we concluded it would be a good idea to gather the various upgrading instructions from around the website into a single unified location that users could use when upgrading Accumulo. This PR is an attempt to collect those instructions from various documents into one location.

### DIFF
--- a/_docs-2-0/administration/upgrading.md
+++ b/_docs-2-0/administration/upgrading.md
@@ -1,0 +1,110 @@
+---
+title: Upgrading Accumulo
+category: administration
+order: 11
+---
+
+## Upgrading from 1.7 to 1.8
+
+Upgrades from 1.7 to 1.8 are possible with little effort as no changes were made at the data layer and RPC changes were made in a backwards-compatible way. The recommended way is to stop Accumulo 1.7, perform the Accumulo upgrade to 1.8, and then start 1.8. Like previous versions, after 1.8 is started on a 1.7 instance, a one-time upgrade will happen by the Master which will prevent a downgrade back to 1.7. Upgrades are still one way. Upgrades from versions prior to 1.7 to 1.8 should follow the below path to 1.7 and then perform the upgrade to 1.8 – direct upgrades to 1.8 for versions other than 1.7 are untested.
+
+Existing configuration files from 1.7 should be compared against the examples provided in 1.8. The 1.7 configuration files should all function with 1.8 code, but you will likely want to include changes found in the 1.8.0 release notes and these release notes for 1.8.1.
+
+For upgrades from prior to 1.7, follow the upgrade instructions to 1.7 first.
+
+## Upgrading from 1.7.x to 1.7.y
+
+The recommended way to upgrade from a prior 1.7.x release is to stop Accumulo, upgrade to 1.7.y and then start 1.7.y.
+
+When upgrading, there is a known issue if the upgrade fails due to outstanding [FATE] operations, see [ACCUMULO-4496] The work around if this situation is encountered:
+
+- Start tservers
+- Start shell
+- Run ```fate print``` to list all
+- If completed, just delete with ```fate delete```
+- Start masters once there are no more fate operations
+
+If any of the FATE operations are not complete, you should rollback the upgrade and troubleshoot completing them with your prior version. When performing an upgrade between major versions, the upgrade is one-way, therefore it is important that you do not have any outstanding FATE operations before starting the upgrade.
+
+## Upgrading from 1.6 to 1.7
+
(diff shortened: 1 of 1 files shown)


_docs-2-0/administration/upgrading.md
```diff
@@ -0,0 +1,110 @@
+---
+title: Upgrading Accumulo
+category: administration
```
mikewalch: Nice. I think it's a good idea this in the documentation. It can be linked to from release notes but it's nice that this will carry forward in the docs with each release for historical purposes.

_docs-2-0/administration/upgrading.md
```diff
@@ -0,0 +1,110 @@
...
+
+After upgrading to 1.7.0, users will notice the addition of a replication table in the accumulo namespace. This table is created and put offline to avoid any additional maintenance if the data-center replication feature is not in use.
+
+Existing configuration files from 1.6 should be compared against the examples provided in 1.7. The 1.6 configuration files should all function with 1.7 code, but you will likely want to include a new file (hadoop-metrics2-accumulo.properties) to enable the new metrics subsystem. Read the section on Hadoop Metrics2 in the Administration chapter of the Accumulo User Manual.
+
+For each of the other new features, new configuration properties exist to support the feature. Refer to the added sections in the User Manual for the feature for information on how to properly configure and use the new functionality.
+Testing
```
mikewalch: `Testing`  can be removed

keith-turner:
Were the sections copied verbatim from the upgrading.md docs?  If so, I am wondering about duplication of documentation.  It would be nice to avoid duplication.  One possible strategy would be to link to the pre 2.0 docs and only have 2.0 upgrade instructions in the manual.  Also remove upgrading.md from 2.0 and modify the 2.0 readme to link to the upgrade instructions on the website.

keith-turner:
Upgrading.md is already gone from master, so there is not duplication for 2.0.0-SNAPSHOT.  So please disregard my comments about duplication.  Need to look at the issue that removed the files in master branch for insights.

https://issues.apache.org/jira/browse/ACCUMULO-4479

_docs-2-0/administration/upgrading.md
ctubbsii: Should the upgrade documentation exist in this centralized docs, or should we focus more on upgrade steps in the release notes, as they apply to that specific release? Either way, there should be a mention in both places. I'm just curious where the details should go. Personally, I like them in the release notes more. I'm just worried that the details will end up spreading across both locations over time.

jmark99: I can understand the preference for release notes, but the conversation started when a user asked how to upgrade from 1.4 to 1.8 and there was some uncertainty as to how easy it was for users to find the upgrade information, especially across multiple version jumps.  The thought was that if there was one central location where the information resided then it would be easier to point users to it and it would also serve as a historical record of upgrade processes. Once the 2.0 docs are 'released' we could update release notes to link to the proper information on this page and allow it to be the official location for upgrade guidance.

keith-turner: > Once the 2.0 docs are 'released' we could update release notes to link to the 

We may also want to consider making the readme.md link to the upgrade docs (instead of having upgrading.md).

ctubbsii: @jmark99 Understood. I'm thinking ahead, to the situation where the 2.0 upgrade section of the docs doesn't fully describe a situation for 2.0.9, because of some bug which specifically affected the 2.0.8 to 2.0.9 upgrade (as an example). Would this get added to the main 2.0 docs, or just called out in the 2.0.9 release notes? Or both? Are we prepared to put all patch-specific upgrade notes in the major-minor release notes, or would we want to describe the overall process, and mention "check the release notes for your version, for any upgrade issues specifically for that version"?

I don't feel strongly about one way or another. I just want to have a plan in place for when we encounter this scenario.

jmark99: I see your point. Given the option, I would lean toward having the overall process described in the upgrade documentation with the stipulation that a user should check the release notes for any specific issues for that particular version. But I'm interested to hear what others would prefer as well.

keith-turner: We could even suggest checking the release notes in the user manual upgrade docs.

ctubbsii: @keith-turner Wait, I'm confused now. I thought that's what we were talking about.

PR #46, opened 2017-12-04
ACCUMULO-4752 Create documentation on improving performance

mikewalch:


_docs-2-0/troubleshooting/peformance.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Read performance
+
+1. Enable [caching] on tables to reduce reads to disk.
+
+1. Enable [bloom filters][bloom-filters] on tables to limit the number of disk lookups.
+
+1. Decrease the [major compaction ratio][compaction] of a table to decrease the number of
+   files per tablet. Less files reduces the latency of reads.
+
```
keith-turner: Adjusting `table.file.compress.blocksize` and `table.file.compress.blocksize.index` can also be helpful.  Lowering `table.file.compress.blocksize` can result in better random seek performance.  However it increases the index size in the file.  If the indexes are too large to fit in cache, this can hinder performance.  Also, as the index size increases the depth of the index tree in each file may increase.  Increasing `table.file.compress.blocksize.index` can reduce the depth of the tree.

_docs-2-0/troubleshooting/peformance.md
```diff
@@ -0,0 +1,39 @@
...
+1. Enable [native maps][native-maps] on tablet servers to prevent Java garbage collection pauses
+   which can slow ingest.
+
+1. [Pre-split new tables][split] to distribute writes across multiple tablet servers.
+
+1. Ingest data using [multiple clients][multi-client] or [bulk ingest][bulk] to increase ingest throughput.
+
+1. Increase the [major compaction ratio][compaction] of a table to limit the number of major compactions
+   which improves ingest peformance.
+
```
keith-turner: Setting `table.file.compress.type=snappy` can increase write spead at the expense of using more disk space.

keith-turner:
@mikewalch the picture looks great.  A metablock contains all of the the root nodes for each locality group.  So this is slightly off in the picture.

mikewalch:
@keith-turner, I updated the picture. Let me know if that works.

PR #48, opened 2017-12-07
Added conditional writer to tour

keith-turner:


tour/conditional-writer.md
```diff
@@ -1,3 +1,119 @@
...
+ConditionalWriter atomically checks conditions on a row and only applies a
+mutation when all are satisfied.
+
+## Exercise
+
+The following code simulates the concurrency situation above.  Because it uses
+a BatchWriter it will lose modifications.
+
+```java
+  static String getAddress(Connector conn, String id) {
```
mikewalch: More imports are needed at top of file.  We could move to wild card imports.  Below worked for me.  However, this should be tested with other exercises.

```java
// Classes you will use along the tour
import java.util.*;
import java.nio.file.*;
import java.util.concurrent.*;
import java.util.function.*;

import org.apache.accumulo.core.client.*;
import org.apache.accumulo.core.client.Scanner;
import org.apache.accumulo.core.data.*;
import org.apache.accumulo.core.security.*;
import org.apache.accumulo.minicluster.MiniAccumuloCluster;
import org.apache.hadoop.io.Text;
```

tour/conditional-writer.md
```diff
@@ -1,3 +1,119 @@
 ---
 title: Conditional Writer
 ---
+
+When read-modify-write operations run concurrently, its possible changes made
+by some operations will be overwritten by others. The following sequence of
+events shows an example of this.
+
+ 1. Thread 0 sets the key `id0001:location:home` to `1007 Mountain Dr, Gotham, New York`
```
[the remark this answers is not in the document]
mikewalch: To add to this, I just think more information needs to be given to set up the exercise so the focus is understanding conditional mutations and not trying to figure out goal of application.

tour/conditional-writer.md
```diff
@@ -1,3 +1,119 @@
...
+Thread  39 attempting change '  1007 Mountain Dr, Gotham, New York  ' -> '  1007 Mountain Dr, Gotham, NY  '
+Thread  37 attempting change '  1007 Mountain Dr, Gotham, New York  ' -> '1007 Mountain Dr, Gotham, New York'
+Final address : '  1007 Mountain Dr, Gotham, NY  '
+```
+
+To fix this example, make the following changes in `setAddress()` to use a
+ConditionalWriter.
+
+ * Call [createConditionalWriter] instead of creating a batch writer
+ * Create a [Condition] for the column 'location:home'.  If `expectedAddr` is not null, then pass it to [setValue].  A condition with no value set means that column is expected to absent.
```
milleruntime: An example on how to create a Condition would be helpful.  I found this tricky never having used it.  I kept looking for a method to set the logic vs explicitly performing the checks in java.

tour/conditional-writer.md
```diff
@@ -1,3 +1,119 @@
...
+ * Replace Mutation with a [ConditionalMutation] and set the condition.
+ * Call [write] passing it the conditional mutation.
+ * Return `true` if [getStatus] from the [Result] returned by [write] is `ACCEPTED`.
```
milleruntime: Referring to the enum here would be helpful: ConditionalWriter.Status.ACCEPTED.  Also some more explanation here would be nice instead of just links to javadoc.  There are a lot of new classes introduced that we haven't used yet.

mikewalch:
This can be merged now.

keith-turner:
merged in 52f812cf45d76f84b3c02fcb15708d77cdc2ea37

PR #61, opened 2018-03-14
Moved BigTable reference to design doc

mikewalch:


index.html
```diff
@@ -9,7 +9,7 @@
...
       <a class="btn btn-success" href="downloads/" role="button"><i class="fa fa-download fa-lg"></i> Download</a>
     </div>
-    <p id="home-description">Apache Accumulo is a key/value store based on the design of Google's <a href="https://research.google.com/archive/bigtable.html">BigTable</a>. Accumulo stores its data in <a href="https://hadoop.apache.org">Apache Hadoop</a>'s HDFS and uses <a href="https://zookeeper.apache.org">Apache Zookeeper</a> for consensus. While many users interact directly with Accumulo, several <a href="/related-projects">open source projects</a> use Accumulo as their underlying store.</p>
+    <p id="home-description">With Apache Accumulo, users can store and manage large data sets across a cluster. Accumulo uses <a href="https://hadoop.apache.org">Apache Hadoop</a>'s HDFS to store its data and <a href="https://zookeeper.apache.org">Apache ZooKeeper</a> for consensus. While many users interact directly with Accumulo, several <a href="/related-projects">open source projects</a> use Accumulo as their underlying store.</p>
```
ctubbsii: This reads like a supporting paragraph. I generally think the description should introduce the project (like a thesis statement), and read like an answer to the question "What IS it?" as in "Apache Accumulo is ....", and only then follow up with what can be done with it, or how it might be used, or what characteristics it might have.

How about: "Apache Accumulo is a distributed sorted storage system for key/value pairs. With it, users can store and manage very large data sets across a cluster. ..."

mikewalch: It is a supporting paragraph in that above it it states `Apache Accumulo® is a sorted, distributed key/value store that provides robust, scalable data storage and retrieval.`

ctubbsii: Ah, I see. I missed that. This is fine, then.

PR #81, opened 2018-05-04
Added search functionality to Accumulo website

mikewalch:
* Search results are limited to blog posts, release notes, and
  documentation for one Accumulo release
* Search is implemented in browser using javascript and lunarjs

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -44,6 +44,7 @@
             <li><a href="{{ site.baseurl }}/related-projects">Related Projects</a></li>
           </ul>
         </li>
+        <li class="nav-link"><a href="{{ site.baseurl }}/search">Search</a></li>
       </ul>
       <ul class="nav navbar-nav navbar-right">
         <li class="dropdown">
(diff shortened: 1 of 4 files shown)


milleruntime:
Oh forgot to say... I pull down your changes and built the website locally.  Looks good.

ctubbsii:
Search button doesn't work when clicked, and menu option messes with the resize to narrow (mobile) screen widths. Could fix the screen resize by replacing "Apache Software Foundation" words with small feather logo as a button (leaving the adjacent drop-down arrow indicator).

ctubbsii:
Saw that you fixed the button and the resize using Apache feather logo. Great work! Thanks!
I do think we probably should avoid the older feather logo in favor of a newer one, if we can find one that looks okay there.

I also noticed a weird thing with the search implementation... it looks like "No results" are returned for "the", but "th" and even "t" return results. This doesn't make much sense to me. If a stop word isn't indexed, then any substring of the stop word should also not be indexed. Meh, whatever. It's pretty good as is.

PR #93, opened 2018-07-09
Created tag for linking to javadocs, properties, and docs

mikewalch:
* Converted most of existing links to use new tags


_config.yml
```diff
@@ -66,10 +68,11 @@ defaults:
       title_prefix: "Accumulo Documentation - "
       version: "2.0"
       latest_release: "2.0.0"
+      javadoc_version: "1.9.1"
```
ctubbsii: Is the expectation that this reflect the current release and the same version as the current docs (after 2.0)?

mikewalch: The `javadoc_version` should be 2.0.0 after 2.0 is released. It's only 1.9.1 so the links work as 2.0.0 is not available on javadoc.io

ctubbsii: That's what I figured. :smiley_cat:

PR #95, opened 2018-07-11
Improved handling of prefix properties

mikewalch:


keith-turner:
It would be useful to add something to readme about prefixes.

PR #100, opened 2018-07-27
Add docs for scan execution hints apache/accumulo#571

keith-turner:


_docs-2-0/administration/scan-executors.md
```diff
@@ -104,11 +104,41 @@ executor.
...
+createtable tex
+config -t tex -s table.scan.dispatcher.opts.heed_hints=true
+```
+
+The {% jlink org.apache.accumulo.core.spi.scan.HintScanPrioritizer %} honors
+hints of the form `priority=<integer>` to prioritize scans, with lower integers
+resulting in a higher priority. The `SimpleScanDispatcher`, which is the
+default dispatcher, supports the `heed_hints` option. By default the
+`SimpleScanDispatcher` ignores hints, but when `heed_hints` is set to `true` it
+will honor hints of the form `executor=<executor name>` when choosing an
```
milleruntime: So the only required property is 'heed_hints'?  Setting a class to tserver.scan.executors.special.prioritizer is optional?

keith-turner: `table.scan.dispatcher` defaults to `org.apache.accumulo.core.spi.scan.SimpleScanDispatcher` which is why only `heed_hints` can be set.   Its probably safer to set it though.

PR #102, opened 2018-08-09
Document exact deletes

keith-turner:


milleruntime:
I think you should mention how this feature will improve performance.

ctubbsii:
This change depends on apache/accumulo#588

keith-turner:
This PR is no longer relevant after closing  apache/accumulo#588 in favor of  apache/accumulo#598

PR #104, opened 2018-08-29
Spelling corrections for Accumulo Website

cjmctague:


_docs-2-0/administration/properties.md
```diff
@@ -35,7 +35,7 @@ Below are properties set in `accumulo-site.xml` or the Accumulo shell that confi
...
+| <a name="general_vfs_context_classpath_prefix" class="prop"></a> **general.vfs.context.classpath.*** | Properties in this category are define a classpath. These properties start  with the category prefix, followed by a context name. The value is a comma separated list of URIs. Supports full regex on filename alone. For example, general.vfs.context.classpath.cx1=hdfs://nn1:9902/mylibdir/*.jar. You can enable post delegation for a context, which will load classes from the context first instead of the parent first. Do this by setting `general.vfs.context.classpath.<name>.delegation=post`, where `<name>` is your context name. If delegation is not specified, it defaults to loading from parent classloader first. |
```
mikewalch: This file is generated from Accumulo code during an Accumulo build. It is periodically copied and committed here. If these changes get committed, there are going to be lost when someone regenerates this file. To change this file, you should edit [Property.java](https://github.com/apache/accumulo/blob/master/core/src/main/java/org/apache/accumulo/core/conf/Property.java) and build Accumulo. This file will be built in `core/target/generated-docs`. It can then build copied and committed here.

cjmctague: Gotcha. Should I just revert these changes and do it on the `Property.java` in accumulo?

mikewalch: Yes. 

You can generate the file, copy it over, and commit it... or just wait for someone else to do it when 2.0 is released.

PR #108, opened 2018-09-13
Add documentation for crypto

milleruntime:


keith-turner:
Might be useful to mention SSL and Kerberose for encrypting data between tserver and client.

_docs-2-0/administration/crypto.md
```diff
@@ -0,0 +1,108 @@
...
+Out of the box, Accumulo provides the `AESCryptoService` for basic encryption needs.  This class provides AES encryption 
+with Galois/Counter Mode (GCM) for RFiles and Cipher Block Chaining (CBC) mode for WALs.  The additional properties 
+below are required by this crypto service to be set using the {% plink instance.crypto.opts.* %} prefix.
+```
+instance.crypto.opts.key.provider=uri
+instance.crypto.opts.key.location=file:///secure/path/to/crypto-key-file
+```
+The first property tells the crypto service how it will get the key encryption key.  The second property tells the service 
+where to find the key.  For now, the only valid values are "uri" and the path to the key file. The key file can be 16 or 32 bytes. 
+Initializing Accumulo after these instance properties are set, will enable on disk encryption across your entire cluster.
```
mikewalch: It would be nice to give users a possible command that creates the key file

_docs-2-0/administration/crypto.md
```diff
@@ -0,0 +1,108 @@
...
+wrapped in at least one other OutputStream.  The `getDecryptionParameters` returns a byte array of anything that will be 
+required to perform decryption. The FileDecrypter only has one method:
+```java
+  InputStream decryptStream(InputStream inputStream) throws CryptoService.CryptoException;
+```
+For more help getting started see {% jlink org.apache.accumulo.core.security.crypto.impl.AESCryptoService %}.
+
+## Things to keep in mind
+
+The on disk encryption configured here is only for R-files and Write Ahead Logs (WALs).  The majority of data in Accumulo
```
mikewalch: Should be `RFile` for consitency

_docs-2-0/administration/crypto.md
milleruntime: @PircDef are there any techniques that you could recommend for creating keys?  If not I could include what I have done for testing... 

echo -n "sixteenbytekey4u" > key-file

PircDef: For testing, I've just been using openssl's random:  

openssl rand -out /path/to/keyfile 32

milleruntime: Ok that is better, I will add it.

PR #109, opened 2018-09-24
Document public API

keith-turner:


pages/api.md
```diff
@@ -0,0 +1,35 @@
...
+title: Public API Definition
+permalink: /api/
+---
+
+Accumulo's public API is composed of all public types in the following
+packages and their sub-packages excluding those named *impl*, *thrift*, or
+*crypto*.
+
+ * org.apache.accumulo.core.client
+ * org.apache.accumulo.core.data
```
mikewalch: It would be nice if these packages were links to the Javadocs

keith-turner: I was not sure if anything was generated at the package level, but poked around and found something.  This would be nice, I will give it a try.

http://accumulo.apache.org/1.9/apidocs/org/apache/accumulo/core/client/package-summary.html

keith-turner: I did this and I really like the result.  Also, I was surprised at how easy it was.  I did not know the custom Ruby plugin for generating javadoc links supported packages.

PR #126, opened 2018-10-29
Created documentation for upgrading to 2.0

mikewalch:
* Updated lexicoder docs in features

### DIFF
--- a/_docs-2/administration/upgrading.md
+++ b/_docs-2/administration/upgrading.md
@@ -4,6 +4,73 @@ category: administration
 order: 7
 ---
 
+## Upgrading from 1.8/9 to 2.0
+
+Follow the steps below to upgrade your Accumulo instance and client to 2.0.
+
+### Upgrade Accumulo instance
+
+**IMPORTANT!** Before upgrading to Accumulo 2.0, you will need to upgrade to Java 8 and Hadoop 3.x.
+
+Upgrading to Accumulo 2.0 is done by stopping Accumulo 1.8/9 and starting Accumulo 2.0.
+
+Before stopping Accumulo 1.8/9, install Accumulo 2.0 and configure it by following the [2.0 install instructions]({% durl getting-started/quick-install %}).
+
+There are several changes to scripts and configuration in 2.0 so be careful when using configuration or automated setup designed for 1.8/9.
+Below are some changes in 2.0 that you should be aware of:
+* `accumulo.properties` has replaced `accumulo-site.xml`. You can either convert `accumulo-site.xml` by hand
+  from XML to properties or use the following Accumulo command.
+  ```
+  accumulo convert-config -x old/accumulo-site.xml -p new/accumulo.properties
+  ```
+* `accumulo-client.properties` has replaced `client.conf`. The [client properties]({% durl configuration/client-properties %})
+  in the new file are different so take care when customizing.
+* `accumulo-cluster` script has replaced the `start-all.sh` & `stop-all.sh` scripts.
+   - Default host files (i.e `masters`, `monitor`, `gc`) are no longer in `conf/` directory of tarball but can be created using `accumulo-cluster create-config`
+   - Tablet server hosts must be listed in a `tservers` file instead of a `slaves` file. To minimize confusion, Accumulo will not start if the old `slaves` file is present.
+* `accumulo-service` script can be used to start/stop Accumulo services (i.e master, tablet server, monitor) on a single node.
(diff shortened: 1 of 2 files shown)


mikewalch:
I wouldn't look at this as final version of upgrade docs.  More instructions can and should be added in follow up PRs.

_docs-2/administration/upgrading.md
```diff
@@ -4,6 +4,73 @@ category: administration
...
+  accumulo convert-config -x old/accumulo-site.xml -p new/accumulo.properties
+  ```
+* `accumulo-client.properties` has replaced `client.conf`. The [client properties]({% durl configuration/client-properties %})
+  in the new file are different so take care when customizing.
+* `accumulo-cluster` script has replaced the `start-all.sh` & `stop-all.sh` scripts.
+   - Default host files (i.e `masters`, `monitor`, `gc`) are no longer in `conf/` directory of tarball but can be created using `accumulo-cluster create-config`
+   - Tablet server hosts must be listed in a `tservers` file instead of a `slaves` file. To minimize confusion, Accumulo will not start if the old `slaves` file is present.
+* `accumulo-service` script can be used to start/stop Accumulo services (i.e master, tablet server, monitor) on a single node.
+    - Can be used even if Accumulo was started using `accumulo-cluster` script.
+* Accumulo 2.0 logging is configured by three files. Any previous logging customization should be added to these files.
```
ctubbsii: This is not necessarily true. The logging depends on the user's use of the log4j system properties in the environment file that sets up the java options when running the process. So, if the user customizes the environment, then these statements about the log config files can become false.

_docs-2/administration/upgrading.md
```diff
@@ -4,6 +4,73 @@ category: administration
...
+    - `log4j-service.properties` for all Accumulo services (except monitor)
+    - `logj4-monitor.properties` for Accumulo monitor
+    - `log4j.properties` for Accumulo clients and commands
+* `accumulo-env.sh` has changed considerably so take care when customizing.
```
ctubbsii: I think it should be mentioned how this file is used. Specifically, its purpose is to construct the `JAVA_OPTS` environment variable, which will be passed as command-line arguments to the `java` command when starting the Accumulo process identified by `$cmd`, as in `java $JAVA_OPTS main.class.for.$cmd`.

Secondarily, it also exports environment, such as `CLASSPATH` which will affect the behavior of `java` when it is executed. Additional environment can also be exported here.

mikewalch: Fixed now

_docs-2/administration/upgrading.md
```diff
@@ -30,12 +30,18 @@ Below are some changes in 2.0 that you should be aware of:
    - Tablet server hosts must be listed in a `tservers` file instead of a `slaves` file. To minimize confusion, Accumulo will not start if the old `slaves` file is present.
 * `accumulo-service` script can be used to start/stop Accumulo services (i.e master, tablet server, monitor) on a single node.
     - Can be used even if Accumulo was started using `accumulo-cluster` script.
-* Accumulo 2.0 logging is configured by three files. Any previous logging customization should be added to these files.
+* `accumulo-env.sh` constructs environment variables (such as `JAVA_OPTS` and `CLASSPATH`) used when running Accumulo processes
+    - This file was used in Accumulo 1.x but has changed signficantly for 2.0
+    - Environment variables (such as `$cmd`, `$bin`, `$conf`) are set before `accumulo-env.sh` is loaded and can be used to customize environment.
+    - The `JAVA_OPTS` variable can be customized to pass command-line arguments to the `java` command that starts Accumulo processes.
```
ctubbsii: This is good... but in my previous review, I used a specific wording, because it's important to understand that `JAVA_OPTS` is special and distinct from the others.

I'm not asking you to use my wording, but I do think it's important to understand that `JAVA_OPTS` is just something we made... and we're explicitly using, whereas things like `CLASSPATH` (and the `MALLOC_*` thing) are not used by us directly, but by other parts of the system. The only thing *we* are specifically asking for (the expected "output" of `accumulo-env.sh`) is `JAVA_OPTS` (although `CLASSPATH` should be set either in the script or before the script is run, otherwise, we won't get very far.

I'm not sure this gives enough attention to `JAVA_OPTS`, since it's really the reason why we're running this script (other exported env, or temporary variables aren't required "output" from this script).

mikewalch: I updated it again with 29a69f0ab4ea01fb2e3ad58520c

does that work?

PR #136, opened 2018-11-29
Update client and server properties

milleruntime:
Update for latest changes to client and server properties.  This should probably wait for the next alpha release, but at least this PR can be a place holder.


milleruntime:
I will revisit this with the 2.0 release.

PR #138, opened 2018-12-10
Add slack to contact us page

mikewalch:


pages/contact-us.md
```diff
@@ -66,6 +66,6 @@ Contributions to Apache Accumulo are welcome! If you are interested, read [how t
 [N_UN]: mailto:notifications-unsubscribe@accumulo.apache.org
 {: .btn .btn-danger .btn-xs }
 
-[accumulo-irc]: irc://chat.freenode.net/accumulo
-[freenode]: https://freenode.net/
+[asf-slack]: https://the-asf.slack.com/
```
mjwall: I think may be a https://the-asf.slack.com/messages/CERNB8NDC is a permalink to the channel

pages/contact-us.md
```diff
@@ -66,6 +66,6 @@ Contributions to Apache Accumulo are welcome! If you are interested, read [how t
...
+[slack-invite]: https://s.apache.org/slack-invite
```
mjwall: I think we should direct users to email dev@a.a.o requesting an invite.

pages/contact-us.md
mikewalch: Added a link to channel

pages/contact-us.md
mikewalch: It would be great if users could join without emailing this list so I kept the invite link but mentioned that they should email dev list if they have trouble joining.

mjwall: Oh, does that work?  Looked like the same link beam is using so I assumed it was setup for them.  If you test and it does work, I approve this PR.  Thanks Mike, I think this is a good idea and hope it gets some use.

ctubbsii: This is a link directly to our channel, if they are already members. Not sure if that's useful: https://the-asf.slack.com/messages/CERNB8NDC

mikewalch: @mjwall, the link worked for me.  @ctubbsii, there is a link to our channel.

PR #144, opened 2019-01-18
Update client documentation

mikewalch:


_docs-2/getting-started/clients.md
```diff
@@ -224,21 +215,37 @@ to efficiently return ranges of consecutive keys and their associated values.
 
 ### Scanner
 
-To retrieve data, Clients use a [Scanner], which acts like an Iterator over
-keys and values. Scanners can be configured to start and stop at particular keys, and
+To retrieve data, create a [Scanner] using [AccumuloClient]. A Scanner acts like an Iterator over
+keys and values in the table.
+
+If a [Scanner] is created without [Authorizations], it use all [Authorizations] granted
```
milleruntime: Typo "it use all"

_docs-2/getting-started/clients.md
```diff
@@ -224,21 +215,37 @@ to efficiently return ranges of consecutive keys and their associated values.
...
+```java
+Scanner s = client.createScanner("table", new Authorizations("public"));
+```
+
+Scanners can be configured to start and stop at particular keys, and
 to return a subset of the columns available.
 
 ```java
-// specify which visibilities we are allowed to see
+// specify which visibilities should be returned
```
milleruntime: This comment is not very clear or accurate.  Perhaps something like "return data with visibilities that match specified auths"

mikewalch: Fixed now

_docs-2/getting-started/clients.md
mikewalch: Fixed

PR #145, opened 2019-01-23
Remove 2.0 warning

milleruntime:
I think we can remove this now that we are advertising the alpha documentation


mikewalch:
How about keeping the warning until 2.0.0 is officially released but updating the text to `This documentation is for an alpha release of Accumulo 2.0.0!` ?

PR #148, opened 2019-01-26
Updates to 2.0 release notes

mikewalch:


_posts/release/2017-09-05-accumulo-2.0.0.md
```diff
@@ -12,8 +12,9 @@ to documentation for new features in addition to issues.
 
 ### New API for creating connections to Accumulo
 
-A fluent API for clients to build `Connector` objects was introduced in [ACCUMULO-4784].
-The new API deprecates `ClientConfiguration` and introduces its own properties
+A fluent API for creating Accumulo clients was introduced in [ACCUMULO-4784] and [#634].
```
keith-turner: I think we need to mention more explicitly what the new client offers that is better than the old API.
  * Can specify everything needed to connect in client config (could not do this in old API).
  * Can configure default settings for things like batch writer in client config (could not do this in old API).
  * Its closeable and explain the static stuff w/ old API
  * Much more intuitive entry point than old API

mikewalch: See my latest commit

PR #156, opened 2019-02-06
Update tour code for 2.0

mikewalch:


src/main/java/tour/Main.java
```diff
@@ -1,38 +1,56 @@
...
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Random;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+import java.util.function.Function;
 
 public class Main {
 
+    static void exercise(MiniAccumuloCluster mac) throws Exception {
```
keith-turner: Do you know if it would be possible to pass the client properties here instead of MiniAccumulo? Seems like that would be cleaner if all the examples do is get the client props from mini

ctubbsii: That would also help examples run against non-mini.

milleruntime:
Closing this since it now has conflicts.

PR #163, opened 2019-03-19
Initial design doc for apache/accumulo#1044

keith-turner:


design/system-snapshot.md
```diff
@@ -0,0 +1,128 @@
...
+## Disclaimer
+
+This is a work in progress design document for a feature that may never be
+implemented.  It represents one possible design for {% ghi 1044 %}, but does
+not have to be the only one.
+
+## Overview
+
+Being able to take snapshots of an entire Accumulo system and roll back to
+snapshot would support many administrative use cases.  This design outlines one
```
mikewalch: could mention some use cases:
* nightly snapshots could be created to recover from major failures (and only lose the current days changes)
* snapshots could created before upgrades to enable rolling back if upgrade fails

keith-turner:
> I like to leverage Git for something like this. Most of everything you have to do is implemented as far as saving storage space and getting an object graph

Using git for operational purposes is useful in some situations, but I am not sure about this one.  A snapshot as outlined in this design is a tree of files in HDFS.  There is a root file with all of the pointers to other files.   The nodes of the tree are immutable, so in that sense git becomes less useful.  Also having some of the nodes of the tree in git and others in hdfs seems like it would increase administrative management overhead.  The data size of the tree could be very large, so it would not make sense to put the entire tree in git.

design/system-snapshot.md
keith-turner: These are mentioned in the issue.  I will mention that.

design/system-snapshot.md
```diff
@@ -0,0 +1,136 @@
...
+
+A user could optionally flush some or all tables before taking a snapshot.
+
+More thought needs to be given to write ahead logs.  This design ignores them
+and only concerns itself with flushed data.
+
+Pausing FATE ops may not be needed.  More design work is needed in the general
+area of FATE ops. Ideally the snapshot operation would be extemely fast.
+Pausing FATE ops could be very slow.  The reason behind pausing is to get a
+consistent view of FATE and the root+metadata tables.
```
milleruntime: Wouldn't it simplify things drastically to take a snapshot of only data at rest?  Ignoring any currently running FATE operations that haven't completed.

keith-turner: I tried to explain this better in the design doc.  Let me know what you think of the updates.  Also while puzzling over how to explain this I realized we only need to pause writes to the FATE data store.  However, we do not need to wait for currently running FATE steps to complete.

design/system-snapshot.md
```diff
@@ -67,12 +71,16 @@ snapshot. The FATE op would do the following :
  1. get snapshot lock in ZK (prevents concurrent snapshot operations)
  1. ensure snapshot name is not in use
  1. pause changing props in ZK
- 1. pause non-snapshot fate ops (let them finish current step, but do not
-    execute next step).   Temporarily stop accepting new FATE ops.
+ 1. pause non-snapshot FATE ops (let running fate ops keep running, but pause
```
mikewalch: could mention something about upgrades and how fate ops cannot be run in upgraded system

keith-turner: I added a section about upgrades.  While writing it I thought of some issues with upgrades and possible solutions.

PR #164, opened 2019-04-02
Update release notes for 1.9.3

milleruntime:
Rough draft of all the changes

### DIFF
--- a/_posts/release/2018-07-18-accumulo-1.9.3.md
+++ b/_posts/release/2018-07-18-accumulo-1.9.3.md
@@ -14,6 +14,16 @@ Users of any previous version of 1.8 or 1.9 are encouraged to upgrade
 
 ## Notable Changes
 
+### Multiple Fixes for Write Ahead Logs
+
+Handle many tablets referencing many WALs [#854] [#860]
+
+Fix improper handling of new WAL failure [#949] [#1005] [#1057]
+
+Fix WAL RecoveryLogReader error on failed file [#961] [#1048]
+
+Fix Unreferenced closed WALs that were never written to [#823] [#845]
+
 ### Fix ArrayOutOfBounds error when new files are created (affects all previous versions)
 
 Accumulo maintains a 1-up counter to keep file names and other identifiers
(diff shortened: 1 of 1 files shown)


_posts/release/2018-07-18-accumulo-1.9.3.md
```diff
@@ -38,10 +48,67 @@ This issue was fixed in pull request [#562]
 
 ### ~Another notable item here~ (affects ~versions~)
 
-More description here. This was fixed in [#NUM].
-
 ## Other Changes
 
+* 07d6346b1 Fix [#994] Consider mutation queue in example config [#1055]
```
ctubbsii: I'm very much against using an "Other changes" section to merely reproduce git history. Anything we put on the website should be curated, added because doing so will be of interest to the user, not simply as a record of the changes.

_posts/release/2018-07-18-accumulo-1.9.3.md
```diff
@@ -14,6 +14,16 @@ Users of any previous version of 1.8 or 1.9 are encouraged to upgrade
 
 ## Notable Changes
 
+### Multiple Fixes for Write Ahead Logs
+
+Handle many tablets referencing many WALs [#854] [#860]
```
ctubbsii: These are a good start, but these should be expanded a bit to explain the changes to the user in a way that helps them understand how it impacts them. Several of these could issues could be listed under the same description, if the high-level overview is the same.

ctubbsii:
@milleruntime, I put "Requested changes" in my review, but I do understand it's a draft, and don't expect all changes to be finalized before merging this. I'm merely highlighting where I think docs could be improved. If I get a chance, I'll try to help with this.

_posts/release/2018-07-18-accumulo-1.9.3.md
milleruntime: I hear ya... I was just putting all of them on here to make it easier for us to "nominate" or bump them up into notable changes.

milleruntime: Also, the link we supplied for a list of other changes for 1.9.2 was only issues and PRs with a label.  It doesn't include commits we made directly, even though these are usually insignificant or fix issues, I wanted us to start with a list of everything and whittle it down.

ctubbsii: If it's helpful, we could do a link like: https://github.com/apache/accumulo/compare/rel/1.9.1...apache:rel/1.9.2

milleruntime: Yeah that might be better for a catch all link at the end.

keith-turner:
@milleruntime  I was looking through the commits and made the following notes.  If you merge this change I can add these or you can add them if you want.

```
### Stop locking during compaction

Compactions were acquiring the tablet lock between each key value. 
This created unnecessary contention with other operations like scan and 
bulk imports.  For {% ghi 1031 %} the the synchronization was removed
by {% ghi 1032 %}. 

### Fixed splitting tablets with files and no data

The split code assumed that if a tablet had files that it had data in
those files.  There are some edge case where this is not true.  Updated
the split code to handle this {% ghi 998 %} {% ghi 999 %} .

### Log when a scan waits a long time for files.

Accumulo has a configurable limit on the max number of files open in a
tserver for all scans.  When too many files are open, scans must wait.
In {% ghi 978 %} and {% ghi 981 %} scans that wait too long for files
now log a message.

### Fixed race condition in table existance check.

The Accumulo client code that checks if tables exists had a race 
condition.  The race was fixed in {% ghi 768 %} and {% ghi 973 %}

### Support running Mini Accumulo using Java 11

Mini Accumulo made some assumptions about classloaders that were no 
longer true in Java 11.  This caused Mini to fail in Java 11.  In 
{% ghi 924 %} Mini was updated to work with Java 11, while still working
with Java 7 and 8.

### Fixed issue with improperly configured Snappy

If snappy was configured for a table and the snappy libraries were not 
available on the system then this could cause minor compactions to hang
forever.  In {% ghi 920 %} and {% ghi 925 %} this was fixed and minor 
compactions will proceed when a different compression is configured.

### Handle bad locality group config.

Improperly configured locality groups could cause a tablet to become
inoperative.  This was fixed in {% ghi 819 %} and {% ghi 840 %}.

### Fixed bulk import race condition.

There was a race condition in bulk import that could result in files
being imported after a bulk import transaction had completed.  In the
worst case these files were already compacted and garbage collected. 
This would cause a tablet to have a reference to a file that did not 
exists.  No data would have been lost, but it would cause scans to fail.
The race was fixed in {% ghi 800 %} and {% ghi 837 %}

TODO document 759 710 669
```

PR #177, opened 2019-04-29
Issue-1136-add-audit-docs add audit log info to docs

DonResnik:
Let me know if there should be more detail on the Audit log section, but it seemed to be about the same level as the other services described.

### DIFF
--- a/_docs-2/configuration/files.md
+++ b/_docs-2/configuration/files.md
@@ -40,7 +40,7 @@ Accumulo processes. See the [quick install] for help with configuring this file.
 ### log4j-service.properties
 
 The {% ghc assemble/conf/log4j-service.properties %} file configures logging for most Accumulo services
-(i.e [Master], [Tablet Server], [Garbage Collector]) except for the Monitor.
+(i.e [Master], [Tablet Server], [Garbage Collector], [Audit]) except for the Monitor.
 
 ### log4j-monitor.properties
 
@@ -87,4 +87,5 @@ to run standby Tracers that can take over if the lead Tracer fails.
 [Tablet Server]: {% durl getting-started/design#tablet-server %}
 [Monitor]: {% durl getting-started/design#monitor %}
 [Tracer]: {% durl getting-started/design#tracer %}
+[Audit]: {% durl getting-started/design#audit %}
 [quick start]: {% durl getting-started/quickstart#configuring-accumulo %}
--- a/_docs-2/configuration/files.md
+++ b/_docs-2/configuration/files.md
@@ -40,7 +40,7 @@ Accumulo processes. See the [quick install] for help with configuring this file.
 ### log4j-service.properties
 
 The {% ghc assemble/conf/log4j-service.properties %} file configures logging for most Accumulo services
-(i.e [Master], [Tablet Server], [Garbage Collector], [Audit]) except for the Monitor.
+(i.e [Master], [Tablet Server], [Garbage Collector]) except for the Monitor.
 
 ### log4j-monitor.properties
 
@@ -87,5 +87,4 @@ to run standby Tracers that can take over if the lead Tracer fails.
 [Tablet Server]: {% durl getting-started/design#tablet-server %}
 [Monitor]: {% durl getting-started/design#monitor %}
 [Tracer]: {% durl getting-started/design#tracer %}
-[Audit]: {% durl getting-started/design#audit %}
 [quick start]: {% durl getting-started/quickstart#configuring-accumulo %}
--- a/_docs-2/getting-started/design.md
+++ b/_docs-2/getting-started/design.md
@@ -89,11 +89,7 @@ forwarding of logs from remote hosts to the Monitor, only one Monitor process sh
 at one time. Leader election will be performed internally to choose the active Monitor.
 
 ### Audit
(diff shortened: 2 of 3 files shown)


_docs-2/getting-started/design.md
```diff
@@ -88,6 +88,13 @@ Multiple Monitors can be run to provide hot-standby support in the face of failu
 forwarding of logs from remote hosts to the Monitor, only one Monitor process should be active
 at one time. Leader election will be performed internally to choose the active Monitor.
 
+### Audit
+Accumulo has a robust audit service that logs most table actions, both successful and 
+failed attempts.  Audit logs are not enabled by default, but can be turned 
+on by editing the conf/log4j-service.properties file.  Audit logs are written to the same 
+location as the master, monitor and tserver logs, unless redirected in the conf/log4j-
+service.properties file.  Audit logs can be configured for the shell in conf/log4j.properties.
```
ctubbsii: This information doesn't accurately represent auditing. Auditing merely uses our normal slf4j logging facilities, but with a named logger (named `org.apache.accumulo.audit`). This logger can be configured by the user in whatever way they normally configure loggers, which would depend on the specific logging implementation they've configured in their deployment.

They are not necessarily enabled or disabled by default, and editing that file does not necessarily enable them, and it is not true that audit logs are written by default to the same location as the other loggers.

The `conf/log4j-service.properties` file is merely an example. Having our documentation specifically written under the assumption that it is the one way to configure logging, conflicts with the reason a lot of the work was done for logging and scripts in 2.0: that reason being to reduce assumptions in our code about the deployment environment, and thereby giving users greater control.

Documentation should reflect the fact that logging is fully configurable by the user. As far as auditing goes, we should indicate the name of the named slf4j logger we send audit logs to, and point to an the log4j configuration file as an example of what a user could do... but we should not place greater emphasis on the example configuration file we ship than it deserves. After all, it won't even have any effect if they don't have log4j on their class path for their slf4j implementation.

_docs-2/configuration/files.md
```diff
@@ -40,7 +40,7 @@ Accumulo processes. See the [quick install] for help with configuring this file.
 ### log4j-service.properties
 
 The {% ghc assemble/conf/log4j-service.properties %} file configures logging for most Accumulo services
-(i.e [Master], [Tablet Server], [Garbage Collector]) except for the Monitor.
+(i.e [Master], [Tablet Server], [Garbage Collector], [Audit]) except for the Monitor.
```
ctubbsii: Auditing is not a service and should not be included in this list.

_docs-2/configuration/files.md
```diff
@@ -87,4 +87,5 @@ to run standby Tracers that can take over if the lead Tracer fails.
 [Tablet Server]: {% durl getting-started/design#tablet-server %}
 [Monitor]: {% durl getting-started/design#monitor %}
 [Tracer]: {% durl getting-started/design#tracer %}
+[Audit]: {% durl getting-started/design#audit %}
```
ctubbsii: Not needed if the reference is removed.

_docs-2/getting-started/design.md
```diff
@@ -88,6 +88,9 @@ Multiple Monitors can be run to provide hot-standby support in the face of failu
 forwarding of logs from remote hosts to the Monitor, only one Monitor process should be active
 at one time. Leader election will be performed internally to choose the active Monitor.
 
+### Audit
+Accumulo has a robust and fully configurable audit logging service that logs most table actions, both successful and failed attempts.  Audit logs are managed through the slf4j named logger `org.apache.accumulo.audit`.  The logger can be added to an existing logging framework or the sample file  `conf/log4j-service.properties` can be used to provide basic audit logging.
+
```
ctubbsii: I think this section looks good, but I don't know if this is the right location for it. Do we have a better location, that talks about logging more generally, where it would be more appropriate?

_docs-2/getting-started/design.md
DonResnik: Chris,

Thanks for the comments, I attempted to capture the main points you stated in my update, but please let me know if there is something missing or wrong.

Maybe we should also update the Configuration Files page at the Log Configuration Files section and add a comment that the log files are samples and that the loggers in those files can be added to any existing log4j logging framework instead of using those files.  Let me know what you think about that, or if you think that should be a separate ticket.

_docs-2/getting-started/design.md
DonResnik: I moved it to in-depth install under the logging section, let me know if that works

DonResnik: Chris, also, sorry about the poorly formatted git commit comments across my branches.  I get it now...won't happen again :-)

ctubbsii:
Merged in da2db3a122892c74be87032351fa1ecba37f6d58

_docs-2/getting-started/design.md
ctubbsii: No worries. It's something everyone does poorly at some time or another. I like to joke that the reason I try to suggest people improve their commit log messages is so that I will continue to hold the record for the longest and worst commit message in Accumulo's git history: https://github.com/apache/accumulo/commit/8851c3578e5b12281a47fdb0b0f7c43516a544d6
(it's pretty bad)

PR #188, opened 2019-08-01
Blog post for upgrading to 2.0

milleruntime:
Some high level bullet points for selling folks on upgrading... We should probably wait and post this after the release.


_posts/blog/2019-08-01-why-upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+for 2.0. Check it out [here][manual]. Users can now quickly [search] the website across all 2.0 documentation.
+
+## Java 8-11 Support
+
+Accumulo 2.0 will work with versions of Java up to 11.  It is not required, but will build and run with Java 11.
+
+## New Crypto
+
+On disk encryption was redone to be more secure and flexible. For an in depth description of how Accumulo 
+does on disk encryption, see the [user manual][crypto].
```
ctubbsii: Should note that these features are still marked "experimental", because full on-disk encryption is still incomplete.

_posts/blog/2019-08-01-why-upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+when upgrading your production cluster so you may be wondering... why should I upgrade?
+
+My top 10 reasons to upgrade. For all changes see the [release notes][rel]
+
+* [Summaries](#summaries)
+* [New Bulk Import](#new-bulk-import)
+* [Simplified Scripts and Config](#simplified-scripts-and-config)
+* [New Monitor](#new-monitor)
+* [New Accumulo Client](#new-accumulo-client)
+* [Hadoop 3](#hadoop-3-support)
```
keith-turner: 1.9 also supports hadoop 3, so its not exclusive to 2.0

_posts/blog/2019-08-01-why-upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+* [Offline creation](#offline-creation)
+* [Search Documentation](#search-documentation)
+* [Java 11 Support](#java-8-11-support)
```
keith-turner: 1.9 also supports Java 11, so its not exclusive to 2.0

ctubbsii: 1.9.3 and earlier did not support Java 11. 1.9.4 will be the first in that series to seriously try to support it. So, currently, it might actually be exclusive to 2.0.

_posts/blog/2019-08-01-why-upgrade.md
milleruntime: Do we have the wording for "experimental" defined somewhere?  If not I think I will add it to the annotation javadoc.  That's the best place i can think of...

ctubbsii: I agree that's the best place.

_posts/blog/2019-08-01-why-upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+See the [GitHub issue][offline].
+
+## Search Documentation
+
+New ability to quickly search documentation on the website. The user manual was completely redone 
+for 2.0. Check it out [here][manual]. Users can now quickly [search] the website across all 2.0 documentation.
+
+## Java 8-11 Support
+
+Accumulo 2.0 will work with versions of Java up to 11.  It is not required, but will build and run with Java 11.
```
ctubbsii: After 2.0.0 (2.0.1, and any 2.1.x and later) will require Java 11 to build, but not to run. But, that's not the case for 2.0.0, which only required Java 8 to build.

_posts/blog/2019-08-01-why-upgrade.md
milleruntime: I could replace the Java or hadoop bullet with something more compelling and unique to 2.0, like the Scan Executors.

milleruntime:
Replaced the mentioned bullet points with more enticing ones.  I think this is ready to go.

_posts/blog/2019-08-12-why-upgrade.md
```diff
@@ -69,12 +73,19 @@ try (AccumuloClient client = Accumulo.newClient()
 ```
 
 As you can see the client is also closable, which gives developers more control over resources.
-See the [Accumulo entry point javadoc][client] for more info. 
+See the [Accumulo entry point javadoc][client].
+
+Key and Mutation have new fluent APIs:
```
keith-turner: Could explicitly mention that these new APIs make it easy to mix types like `String` and `byte[]`.  This was cumbersome in previous APIs

_posts/blog/2019-08-12-why-upgrade.md
```diff
@@ -0,0 +1,138 @@
+---
+title: "Top 10 Reasons to Upgrade"
+author: Mike Miller
+reviewers: Keith Turner, Christopher Tubbs
```
ctubbsii: I don't want my name here. I also don't think any individual names should be here. I think individual names here are too similar to `@author` tags in code, promoting the idea of individuals over community, which is something that the ASF expressly decided against many years ago. The commit logs show the authors and the mailing list show reviewer feedback. That should be sufficient if somebody wants to go back and review that information. For the purpose of the presentation on the website, the content should reflect the community as a whole.

Also, this "reviewers" line is just plain weird to me. It's likely that others reviewed the page, too, but didn't find have any suggestions for changes. Does their passive concurrence count for less?

I don't know... perhaps these should be discussed on the dev@ list, and perhaps the reviewers tag can be addressed separately from the author tag. In either case, I'm not comfortable with my name being added here.

milleruntime: I only did those tags because thats what we did here: https://accumulo.apache.org/blog/2016/12/19/running-on-fedora-25.html

milleruntime: Feel free to remove the authors and reviewers from posts... doesn't matter to me.

ctubbsii: I'm still forming my opinions on this one, so I won't take any action until I have a stronger opinion. At the very least, I was surprised to see my name added when I didn't add it. But, I don't blame anybody for following the precedent that was set.

PR #189, opened 2019-08-04
Update website for the 2.0.0 release

ctubbsii:


pages/api.md
```diff
@@ -27,8 +28,17 @@ The following regex matches imports that are *not* Accumulo public API. This
...
+```regex
+import\s+org\.apache\.accumulo\.(.*\.(impl|thrift|crypto)\..*|(?!(core\.(client|data|security)|minicluster)\.).*)
 ```
-import\s+org\.apache\.accumulo\.(.*\.(impl|thrift|crypto)\..*|(?!core|minicluster).*|core\.(?!client|data|security).*)
+
+For 2.0 and later, this can be simplified, because sub-packages not intended
+for public API were relocated, and also altered to include the new MapReduce module:
+
+```regex
+import\s+org\.apache\.accumulo\.(?!(core\.(client|data|security)|minicluster|hadoop)\.).*
```
keith-turner: Did you test this regex?  I wrote a standalone java program to test the exisitng regex with different strings.

pages/downloads.md
```diff
@@ -56,25 +55,25 @@ Be sure to verify your downloads by these [procedures][VERIFY_PROCEDURES] using
...
-| **Source** | [accumulo-2.0.0-alpha-2-src.tar.gz][SRC_20] | [ASC][ASC_SRC_20] | [SHA][SHA_SRC_20] |
+| **Binary** | [accumulo-2.0.0-bin.tar.gz][BIN_20] | [ASC][ASC_BIN_20] | [SHA][SHA_BIN_20] |
+| **Source** | [accumulo-2.0.0-src.tar.gz][SRC_20] | [ASC][ASC_SRC_20] | [SHA][SHA_SRC_20] |
 
 #### 2.0 Documentation
 * [README][README_20]
 * [Online Documentation][MANUAL_20]
 * [Java API][JAVADOC_20]
 
-### 1.9.3 **latest**{: .label .label-primary }
```
keith-turner: What does **maintenance** mean?

ctubbsii: Not sure, precisely. I was thinking it was a maintenance release for 1.9. Most importantly, it wasn't "latest". I changed it to "previous", which I think is only marginally better. I could remove the label entirely, but then there's not really a visual marker delineating one release from the next one on the page.

pages/api.md
ctubbsii: I did test it, with https://www.regextester.com, with both the JS engine and the PCRE one. These regexes don't employ anything that would differ substantially from Java's implementation, but I will test it on Java to be sure, before merging.

ctubbsii: Yep, tested with Java. Works the same as JavaScript and PCRE.

PR #197, opened 2019-10-16
revise grammar

rcarterjr:
I came across a minor grammar error while reading the docs.

### DIFF
--- a/1.3/user_manual/Writing_Accumulo_Clients.md
+++ b/1.3/user_manual/Writing_Accumulo_Clients.md
@@ -25,7 +25,7 @@ All clients must first identify the Accumulo instance to which they will be comm
 
 ## <a id="Writing_Data"></a> Writing Data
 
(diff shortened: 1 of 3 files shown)


ctubbsii:
For future reference, "data are" is grammatically correct, even though it doesn't seem like it: https://blog.apastyle.org/apastyle/2012/07/data-is-or-data-are.html
("data" is plural, "datum" is singular)

However, in this case, the incorrect form is more familiar, and it probably doesn't matter much.

AdamAce38:
https://repo1.maven.org/maven2/com/google/

AdamAce38:
https://ci.jenkins.io/job/Core/job/jenkins/job/PR-8814/3/

AdamAce38:
distfiles.m❔❔❔ort❔.org   1PASSWORD

AdamAce38:
https://distfiles.macports.org

PR #202, opened 2019-10-25
Update for 2.0

milleruntime:
One solution for getting the tour to work with MAC for 2.0

### DIFF
--- a/src/main/java/tour/Main.java
+++ b/src/main/java/tour/Main.java
@@ -47,8 +47,8 @@ public class Main {
 
         mac.start();
 
-	try (AccumuloClient client = mac.createAccumuloClient("root", new PasswordToken("tourpass"))) {
-            exercise(client);
+        try (AccumuloClient client = mac.createAccumuloClient("root", new PasswordToken("tourpass"))) {
+                exercise(client);
         }
         mac.stop();
     }
--- a/src/main/java/tour/Main.java
+++ b/src/main/java/tour/Main.java
@@ -3,12 +3,15 @@ package tour;
 // Classes you will use along the tour
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Properties;
 import java.util.Random;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Future;
 import java.util.function.Function;
+
+import org.apache.accumulo.core.client.Accumulo;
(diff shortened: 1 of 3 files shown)


keith-turner:
I tried something locally that seemed to work and I think is simpler.  I used [mvn exec:exec to run java](https://www.mojohaus.org/exec-maven-plugin/examples/example-exec-for-java-programs.html) instead of `mvn exec:java`.

So  I changed the pom to the following

```xml
      <plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>exec-maven-plugin</artifactId>
        <version>1.5.0</version>
        <configuration>
          <executable>java</executable>
          <arguments>
            <argument>-classpath</argument>
            <classpath/>
            <argument>tour.Main</argument>
          </arguments>
          <cleanupDaemonThreads>false</cleanupDaemonThreads>
        </configuration>
      </plugin>
```

And ran the tour using `mvn -q clean compile exec:exec` instead of `mvn -q clean compile exec:java`.   Doing this avoided using the special classloader magic that `exec:java` seemed to do and just set the classpath on the java command line.

keith-turner:
The `<cleanupDaemonThreads>` config option may only be used for `exec:java` and may be ignored for `exec:exec`.  I just copied this over, but did not research it.

milleruntime:
This worked for me too, thanks.  

I was going down this path at one point but was following steps to [use a special classpath variable](https://www.mojohaus.org/exec-maven-plugin/examples/example-exec-for-java-programs.html) 
which did not work for me.

src/main/java/tour/Main.java
```diff
@@ -42,14 +43,17 @@ public static void main(String[] args) throws Exception {
         System.out.println("Running the Accumulo tour. Having fun yet?");
 
         Path tempDir = Files.createTempDirectory(Paths.get("target"), "mac");
-        MiniAccumuloCluster mac = new MiniAccumuloCluster(tempDir.toFile(), "tourguide");
+        MiniAccumuloCluster mac = new MiniAccumuloCluster(tempDir.toFile(), "tourpass");
 
         mac.start();
-        exercise(mac);
+
+	try (AccumuloClient client = mac.createAccumuloClient("root", new PasswordToken("tourpass"))) {
```
ctubbsii: nit; replace tabs with spaces.
```suggestion
	try (AccumuloClient client = mac.createAccumuloClient("root", new PasswordToken("tourpass"))) {
```

PR #206, opened 2019-11-05
Checkstyle import control blog post. Fixes #149

### LINKED ISSUE #149: Update API documentation checkstyle recommendation
Consider updating the website API docs to suggest using the approach used in apache/accumulo-testing#52 instead of the current regex.

milleruntime:


_posts/blog/2019-11-04-checkstyle-import-control.md
```diff
@@ -0,0 +1,93 @@
...
+
+```xml
+<plugin>
+    <!-- This was added to ensure project only uses public API -->
+    <groupId>org.apache.maven.plugins</groupId>
+    <artifactId>maven-checkstyle-plugin</artifactId>
+    <version>3.1.0</version>
+    <configuration>
+      <configLocation>checkstyle.xml</configLocation>
+    </configuration>
```
ctubbsii: There are different implications for putting this config here rather than in the execution section below. This config will be used even for manual runs of the plugin. I'm not sure that's the behavior we want to encourage people to use. Maybe move this config to the execution section below.

_posts/blog/2019-11-04-checkstyle-import-control.md
```diff
@@ -0,0 +1,93 @@
...
+    <dependencies>
+      <dependency>
+        <groupId>com.puppycrawl.tools</groupId>
+        <artifactId>checkstyle</artifactId>
+        <version>8.23</version>
+      </dependency>
+    </dependencies>
```
ctubbsii: This section is optional and could be removed in the example, unless we need to override the version of checkstyle to get a specific version for this functionality to work. Otherwise, most people are probably just going to use the version used by the plugin. Removing this makes it more likely that the user will have a good experience here, because some versions of the checkstyle dependency may not work as well as those tested by the plugin developers.

_posts/blog/2019-11-04-checkstyle-import-control.md
```diff
@@ -0,0 +1,93 @@
...
+## Import Control Configuration
+
+Create the second file specified above, ```import-control.xml``` and copy the configuration below:
+```xml
+<!DOCTYPE import-control PUBLIC
+    "-//Checkstyle//DTD ImportControl Configuration 1.4//EN"
+    "https://checkstyle.org/dtds/import_control_1_4.dtd">
+
+<!-- This checkstyle rule is configured to ensure only use of Accumulo API -->
+<import-control pkg="org.apache.accumulo.testing" strategyOnMismatch="allowed">
```
ctubbsii: Is this `pkg` value correct for users? It looks like this will only enforce import control over classes in our own testing code. It would be useful to explain to the reader what they must do to apply it to their own code.

And, is it worth mentioning what the `strategyOnMismatch` does? I'm kind of curious myself.

_posts/blog/2019-11-04-checkstyle-import-control.md
```diff
@@ -0,0 +1,93 @@
+---
+title: "Checking API use"
+---
+
+Accumulo follows [SemVer] across versions with the declaration of a public API.  Code not in the public API should be
+considered unstable, at risk of changing between versions.  The packages included in the public API are [listed on the website][api]
+but may not always be considered when developing using Accumulo code.  This blog post explains how to setup a Maven project
```
keith-turner:
```suggestion
but may not be considered when an Accumulo user writes code.  This blog post explains how to make Maven
```

_posts/blog/2019-11-04-checkstyle-import-control.md
```diff
@@ -0,0 +1,93 @@
+---
+title: "Checking API use"
+---
+
+Accumulo follows [SemVer] across versions with the declaration of a public API.  Code not in the public API should be
+considered unstable, at risk of changing between versions.  The packages included in the public API are [listed on the website][api]
```
keith-turner:
```suggestion
considered unstable, at risk of changing between versions.  The public API packages API are [listed on the website][api]
```

_posts/blog/2019-11-04-checkstyle-import-control.md
```diff
@@ -0,0 +1,93 @@
...
+but may not always be considered when developing using Accumulo code.  This blog post explains how to setup a Maven project
+to automatically detect when Accumulo 2.0 code used in the project is outside of the public API.
```
keith-turner:
```suggestion
automatically detect usage of Accumulo 2.0 code outside the public API.
```

_posts/blog/2019-11-04-checkstyle-import-control.md
keith-turner: I am learning more about Maven from these comments.

_posts/blog/2019-11-04-checkstyle-import-control.md
```diff
@@ -0,0 +1,93 @@
...
+    <!-- API packages -->
+    <allow pkg="org.apache.accumulo.core.client"/>
+    <allow pkg="org.apache.accumulo.core.data"/>
+    <allow pkg="org.apache.accumulo.core.security"/>
+    <allow pkg="org.apache.accumulo.core.iterators"/>
+    <allow pkg="org.apache.accumulo.minicluster"/>
+    <allow pkg="org.apache.accumulo.hadoop.mapreduce"/>
+
+    <!-- SPI package -->
+    <allow pkg="org.apache.accumulo.core.spi"/>
```
keith-turner: SPI is not currently public API.. not sure if we should include it in this post.

_posts/blog/2019-11-04-checkstyle-import-control.md
```diff
@@ -0,0 +1,93 @@
+---
+title: "Checking API use"
+---
+
+Accumulo follows [SemVer] across versions with the declaration of a public API.  Code not in the public API should be
+considered unstable, at risk of changing between versions.  The packages included in the public API are [listed on the website][api]
+but may not always be considered when developing using Accumulo code.  This blog post explains how to setup a Maven project
+to automatically detect when Accumulo 2.0 code used in the project is outside of the public API.
+
```
keith-turner: I think it would be really useful to add a sentence like the following.  This info is very important.

```
The techniques described in this blog post only work for Accumulo 2.0 and later.  Do not use these with 1.X versions because those version have `impl` packages inside API packages that must be excluded.  The needed exclusions for 1.X are not supported by import control.
```

_posts/blog/2019-11-04-checkstyle-import-control.md
milleruntime: > This config will be used even for manual runs of the plugin.

What is wrong with that?  I would assume if this gets added to the checkstyle plugin, it would always run when checkstyle runs.

_posts/blog/2019-11-04-checkstyle-import-control.md
milleruntime: You are correct, that should be whatever the package of their project is.

_posts/blog/2019-11-04-checkstyle-import-control.md
milleruntime: Good point.  I will remove it.

_posts/blog/2019-11-04-checkstyle-import-control.md
milleruntime: I assume you meant: "The public API packages are [listed on the website][api]"

_posts/blog/2019-11-04-checkstyle-import-control.md
milleruntime: Yeah I wasn't sure either.  We don't list it on the site so I'll remove it.

_posts/blog/2019-11-04-checkstyle-import-control.md
milleruntime: You are right, but I don't think we need to try to explain why.  I added a shortened version of this warning.

_posts/blog/2019-11-04-checkstyle-import-control.md
keith-turner: yes

_posts/blog/2019-11-04-checkstyle-import-control.md
milleruntime: > And, is it worth mentioning what the `strategyOnMismatch` does? I'm kind of curious myself.

I don't remember exactly why this was needed so I didn't mention it.

_posts/blog/2019-11-04-checkstyle-import-control.md
ctubbsii: > What is wrong with that?

It's not necessarily "wrong"... but it can make it more difficult to run the checkstyle plugin with different configuration in separate executions. Moving the `<configuration/>` element and its contents to inside the `<execution/>` element will make it so that execution is specifically configured to do the import control checks, but it wouldn't affect any other executions of the checkstyle plugin the user may already have configured (or might want to configure later), such as additional checks in a separate execution defined in a profile, or executions run manually on the command-line, or independent executions in sibling modules in a multi-module project. In the case of sibling modules, it might be useful for the user to have separate import-control configs, each with their own `<import-control pkg="my.package.name.example">` settings for that specific module, rather than have one massive config file for all modules.

_posts/blog/2019-11-04-checkstyle-import-control.md
```diff
@@ -0,0 +1,86 @@
...
+    <allow pkg="org.apache.accumulo.core.iterators"/>
+    <allow pkg="org.apache.accumulo.minicluster"/>
+    <allow pkg="org.apache.accumulo.hadoop.mapreduce"/>
+
+    <!-- disallow everything else coming from accumulo -->
+    <disallow pkg="org.apache.accumulo"/>
+</import-control>
+```
+This file configures the ImportControl module to only allow packages that are declared public API.
+
```
keith-turner: Could add a section like :

## Hold the line

Adding this to an existing project may expose usage of non public Accumulo API's.  It may take more time than is available to fix those at first, but do not let this discourage adding this plugin.  One possible way to proceed is to allow the currently used non-public APIs in a commented section of `import-control.xml` noting these are temporarily allowed until they can be removed.  This strategy prevents new usages of non-public APIs while allowing time to work on fixing the current usages of non public APIs.

_posts/blog/2019-11-04-checkstyle-import-control.md
```diff
@@ -0,0 +1,86 @@
...
+The techniques described in this blog post only work for Accumulo 2.0 and later.  Do not use with 1.X versions.
+
+## Checkstyle Plugin
+
+First add the checkstyle Maven plugin to your pom.
+
+```xml
+<plugin>
+    <!-- This was added to ensure project only uses public API -->
```
keith-turner:
```suggestion
    <!-- This was added to ensure project only uses Accumulo's public API -->
```

_posts/blog/2019-11-04-checkstyle-import-control.md
milleruntime: I think this strategy is needed because the module is so explicit.  Without allowing a mismatch, any import not allowed will get flagged, like even java imports.

PR #208, opened 2019-11-12
Update blog to include openssl path to JAVA_OPTS in accumulo-env.sh

karthick-rn:
Hi @keith-turner, Based on your comment earlier I have updated the blog to include `-Dorg.wildfly.openssl.path` to `JAVA_OPTS` in `accumulo-env.sh`.  This does not throw any errors during start/stop accumulo and also I have tested running live ingest with this setting for the past few hours and all looks good. Let me know for any questions/comments? Thanks & sorry this took a bit of time as I was on holiday :)


keith-turner:
Thanks for updating the asf-site branch @billierinaldi .  I forgot to do that.

PR #210, opened 2019-11-15
Change seeked to sought

phrocker:
Only updated in docs_2

### DIFF
--- a/_docs-2/apidocs/org/apache/accumulo/core/data/Condition.html
+++ b/_docs-2/apidocs/org/apache/accumulo/core/data/Condition.html
@@ -611,7 +611,7 @@ extends <a href="https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html
 <div class="block">Set iterators to use when reading the columns value. These iterators will be applied in
  addition to the iterators configured for the table. Using iterators its possible to test other
  conditions, besides equality and absence, like less than. On the server side the iterators will
- be seeked using a range that covers only the family, qualifier, and visibility (if the
+ be sought using a range that covers only the family, qualifier, and visibility (if the
  timestamp is set then it will be used to narrow the range). Value equality will be tested using
  the first entry returned by the iterator stack.</div>
 <dl>
--- a/_docs-2/getting-started/table_design.md
+++ b/_docs-2/getting-started/table_design.md
@@ -85,7 +85,7 @@ date range.
 Since Keys in Accumulo are sorted lexicographically by default, it's often useful to encode
 common data types into a byte format in which their sort order corresponds to the sort order
 in their native form. An example of this is encoding dates and numerical data so that they can
-be better seeked or searched in ranges.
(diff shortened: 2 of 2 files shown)


phrocker:
This may seem trite, but I never merge a change in docs without review, no matter how small.

_docs-2/apidocs/org/apache/accumulo/core/data/Condition.html
```diff
@@ -611,7 +611,7 @@ <h4>setIterators</h4>
 <div class="block">Set iterators to use when reading the columns value. These iterators will be applied in
  addition to the iterators configured for the table. Using iterators its possible to test other
  conditions, besides equality and absence, like less than. On the server side the iterators will
- be seeked using a range that covers only the family, qualifier, and visibility (if the
+ be sought using a range that covers only the family, qualifier, and visibility (if the
```
keith-turner: This corrects the english, but I feel its not as descriptive of what it going on in the tserver.  What do you think about the following

```
On the server side `seek()` is called on the iterators using a range ...
```

I think this may be generated javadoc.  Could update it in the source code also.

keith-turner: Ugh, just realized thats javadoc so can't use markdown, so maybe :

```
On the server side {@code seek()} is called on the iterators using a range ...
```

keith-turner: Actually there may be an advantage to it being javadoc, could actually link to the seek function.

phrocker: Ohhh that's a good point I see it's generated by the javadoc. Thanks!

_docs-2/getting-started/table_design.md
```diff
@@ -85,7 +85,7 @@ date range.
 Since Keys in Accumulo are sorted lexicographically by default, it's often useful to encode
 common data types into a byte format in which their sort order corresponds to the sort order
 in their native form. An example of this is encoding dates and numerical data so that they can
-be better seeked or searched in ranges.
+be better sought or searched in ranges.
```
milleruntime:
```suggestion
perform better during a seek or range search.
```

_docs-2/apidocs/org/apache/accumulo/core/data/Condition.html
ctubbsii: Since this is generated, it's probably best to just fix the wording in the javadoc. I'm going to close this PR, with @milleruntime 's suggestion for `table_design.md` so the branch that GitHub automatically created when the file was edited from the web UI can be removed.

_docs-2/getting-started/table_design.md
ctubbsii: Merged in 24b0eb6d124e844acad27f6222605af7171aea7a

PR #215, opened 2019-12-22
Update licenses.html

Graceas:
Fixed Spelling


ctubbsii:
This page is generated automatically from maven-site-plugin tooling. Instead of fixing this here, I suggest fixing the broken tooling. When the tooling is fixed, I'd be happy to regenerate the site using the updated tooling. But, I don't want to merge this, because the typo will just keep getting reintroduced every time we generate this part of the site. If you need help identifying the specific tooling code containing the typo, I may be able to help with that after the new year.

PR #216, opened 2019-12-23
Proxy client code updates

mikewalch:
* Moved code from proxy docs on website to the accumulo-proxy repo
* Updated names for client scripts


mikewalch:
Related to apache/accumulo-proxy#17

PR #223, opened 2020-02-28
Adds a design document for external compactions

keith-turner:
In a conversation with @ctubbsii one day he asked me if a distributed queue
could be used instead of the approach proposed in apache/accumulo#1451.  A few weeks later I
brought this up in a discussion with @billierinaldi and we worked out a
possible way to do this on a whiteboard.  This document captures that.

### DIFF
--- a/design/external-compactions.md
+++ b/design/external-compactions.md
@@ -1,0 +1,245 @@
+---
+title: External Compaction Design
+---
+
+## Definitions
+
+ * *External compaction:* compactions of a tablet execute in a process other than
+   the tablet server hosting that tablet.
+
+ * *Internal compaction:* compactions of a tablet execute in the tablet server
+   process hosting the tablet.
+
+ * *Compactor:* Accumulo process that runs external compactions.
+
+## Introduction
+
+Currently, Accumulo only supports internal compactions.  This can lead to
+uneven load on a cluster.  For example, a few tablet servers could have many
+tablets to compact while many tablet servers are idle.  If Accumulo supported
+external compactions, then compaction work could evenly spread across a
+cluster.
+
+Compactors could start with a command like:
+
+```
+  accumulo compactor <queue>
+```
+
+This would start a process that looks for compactions on the specified
+distributed queue and executes them.  The command could easily run in a docker
+container in something like Kubernetes.  A compactor would need to continually
+do the following
+
+ * Find tablets with files to compact in the queue
+ * Reserve files/work unit
+ * Compact files
+ * Commit compaction
+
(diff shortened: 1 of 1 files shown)


design/external-compactions.md
```diff
@@ -0,0 +1,245 @@
...
+## Queues
+
+A distributed queue for external compactions needs to support the following operations.
+
+ * Adding and removing compaction work for a tablet
+ * Prioritizing compaction work for a tablet
+ * Efficiently finding work
+
+One possibility to implement a distributed queue is a section in Accumulo’s
+metadata table for each queue.  Giving each queue a unique row prefix within
```
milleruntime: Do we want external processes (possibly many) to constantly be scanning the metadata table?  Would it be better to have a separate table for the compaction queue?

milleruntime:
I was wondering what your thoughts were with this design and merging minor compactions?  I would think their benefits would be minimized and removing them would greatly reduce complexity.

design/external-compactions.md
keith-turner: A goal of the design is to avoid that.  When a compactor is running a compaction it will not be scanning the metadata table.  When it its not running a compaction and searching there are a few things that should help reduce the load on the metadata table.  First whenever a compactor finds nothing or has a reservation collision, it will do exponential backoff before looking again.  Second, on purpose of the bins is to avoid reservation collisions (those could cause a lot of metadata activity).   

Do you think it would be useful to add a section about this?

keith-turner:
> I was wondering what your thoughts were with this design and merging minor compactions? I would think their benefits would be minimized and removing them would greatly reduce complexity.

I had not considered that, but I think removing them may be a good idea.  A smart selection algorithm plus multiple compaction queues (like this [proposal](https://gist.github.com/keith-turner/16125790c6ff0d86c67795a08d2c057f)) could remove the need for merging minor compactions.  I think one would want compactions for small files to run on an internal queue for responsiveness.  Compactions of larger files could run on an external queue.

design/external-compactions.md
milleruntime: No your write up is solid and I did see your comments in the others sections about collisions.  I was thinking more of scale or performance and preventing users shooting themselves in the foot by crippling the metadata with too many Compactors or poorly configured Compactors.

ctubbsii:
I agree with the above statements about smart selection, multiple queues, and/or external compactions, removing the need for merging minor compactions. I think merging minor compactions provide too little value to justify their complexity, especially in light of these better alternatives. I'd be happy if they went away.

PR #226, opened 2020-04-02
Doc update for ZooKeeper stat command config for Monitor Overview

EdColeman:
Running latest with uno noticed that Monitor Overview page was not showing Zookeeper info.  This
adds documentation to enable stat command (required with zookeeper 3.5.3+)

### DIFF
--- a/_docs-2/configuration/overview.md
+++ b/_docs-2/configuration/overview.md
@@ -85,12 +85,14 @@ client.tableOperations().setProperty("mytable", "table.durability", "log");
 ```
 
 ### Zookeeper Considerations
+
 Any [server properties] that are set in Zookeeper should consider the limitations of Zookeeper itself with respect to the
 number of nodes and the size of the node data. Custom table properties and options for Iterators configured on tables
 are two areas in which there aren't any fail safes built into the API that can prevent the user from making this mistake.
 
 While these properties have the ability to add some much needed dynamic configuration tools, use cases which might fall
 into these warnings should be reconsidered.
+
 ## Viewing Server Configuration
 
 Accumulo's current configuration can be viewed in the shell using the `config` command.
@@ -131,4 +133,3 @@ default  | table.failures.ignore ..................... | false
 [table.* properties]: {% purl table.\* %}
 [accumulo-client.properties]: {% durl configuration/files#accumulo-clientproperties %}
 [accumulo.properties]: {% durl configuration/files#accumuloproperties %}
-[Zookeeper Administration Documentation]: https://zookeeper.apache.org/doc/r3.5.7/zookeeperAdmin.html 
--- a/_docs-2/configuration/overview.md
+++ b/_docs-2/configuration/overview.md
@@ -85,9 +85,6 @@ client.tableOperations().setProperty("mytable", "table.durability", "log");
 ```
 
 ### Zookeeper Considerations
-
-#### Storing Accumulo Server Properties
-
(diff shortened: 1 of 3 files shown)


_docs-2/configuration/overview.md
```diff
@@ -86,13 +86,26 @@ client.tableOperations().setProperty("mytable", "table.durability", "log");
...
+
+Starting with Zookeeper version 3.5.3, the Four Letter Words commands must be enabled with the 
+zookeeper property: 4lw.commands.whitelist.  The monitor uses the stat command to provide zookeeper server(s),
+mode and number of clients.  At a minimum, add the following to your zookeeper configuration (zoo.cfg):
+```
+4lw.commands.whitelist=stat
+```
+See [Zookeeper Administration Documentation] for your zookeeper version for other four-letter-word 
+command whitelist options.
+
```
ctubbsii, line 108: I don't think this section is necessary at all. It should be sufficient to link directly to ZK docs on the matter, rather than duplicate the information here, so we don't become a configuration guide for ZK itself. See my other review comments for more.

```suggestion
```

_docs-2/configuration/overview.md
```diff
@@ -86,13 +86,26 @@ client.tableOperations().setProperty("mytable", "table.durability", "log");
 
 ### Zookeeper Considerations
 
+#### Storing Accumulo Server Properties
+
```
ctubbsii, line 90: This entire page is content describing how Accumulo configuration works. I don't think ZooKeeper configuration tips should be put here, as it is out of scope for this page.

Also, I don't think it's necessary to modify this page at all (see other review comments).

```suggestion
```

ctubbsii:
> Running latest with uno noticed that Monitor Overview page was not showing Zookeeper info. This
> adds documentation to enable stat command (required with zookeeper 3.5.3+)

I had thought Uno was adding this already. It should soon (See apache/fluo-uno#245).

_docs-2/configuration/overview.md
EdColeman: I think your points are valid - the issue that I had in creating this was were would I go, if on an upgrade, the monitor stopped showing the ZooKeeper info.  The configuration section seemed like a place I might go first - the monitor pages seems to provide more general info that existing users may skip - but, it is a monitor issue, so, maybe there?  And rather than duplicate info, the link.

The essence I what I wanted to convey is that the monitor relies on the stat command.  Other ZooKeeper operations are using a connector and will continue to work - but the monitor is using an alternate connection path and that may not be obvious users.  With the zookeeper white list changes, they need to do update their existing  zookeeper configuration. 

Fixing uno helps new instance experiences - what I hoped to convey was that exiting users may need adjustments - and those might not have been obvious

ctubbsii: The main intent I have in my feedback is to keep the documentation concise. We already have so much documentation, it's overwhelming and critical information gets buried in pages and pages of superfluous explanatory text. Every addition to the docs that focuses on one doc deficiency at a time risks making the situation slightly worse, which builds up over time. I just want to avoid that and try to think about the impact on the totality of the documentation by keeping the additions concise.

> The essence I what I wanted to convey is that the monitor relies on the stat command.

Right. I think that's the best thing to focus on. I just want it to be concise.

As for where this should live, I think your original "Note" is fine (did my proposed changes to that satisfy your intent in a concise way?). It could also live in the "Troubleshooting" section.

PR #232, opened 2020-06-26
Update Compaction documentation apache/accumulo#1613

keith-turner:
This should not be merged until after 2.1.0 is released.

### DIFF
--- a/_docs-2/administration/compaction.md
+++ b/_docs-2/administration/compaction.md
@@ -1,0 +1,118 @@
+---
+title: Compactions
+category: administration
+order: 6
+---
+
+In Accumulo each tablet has a list of files associated with it.  As data is
+written to Accumulo it is buffered in memory. The data buffered in memory is
+eventually written to files in DFS on a per tablet basis. Files can also be
+added to tablets directly by bulk import. In the background tablet servers run
+major compactions to merge multiple files into one. The tablet server has to
+decide which tablets to compact and which files within a tablet to compact.
+
+Within each tablet server there are one or more user configurable Comapction
+Services that compact tablets.  Each Accumulo table has a user configurable
+Compaction Dispatcher that decides which compaction services that table will
+use.  Accumulo generates metrics for each compaction service which enable users
+to adjust compaction service settings based on actual activity.
+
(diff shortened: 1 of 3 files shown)


_docs-2/administration/compaction.md
```diff
@@ -0,0 +1,119 @@
...
+tablet means more higher query latency. So adjusting this ratio is a trade off
+between ingest and query performance.
+
+When CR=1.0 this will result in a goal of a single per file tablet, but the
+amount of work is O(N<sup>2</sup>) so 1.0 should be used with caution.  For
+example if a tablet has a 1G file and 1M file is added, then a compaction of
+the 1G and 1M file would be queued. 
+
+Compaction services and dispatchers were introduced in Accumulo 2.1, so much
+of this documentation only applies to Accumulo 2.1 and later.
```
brianloss, line 38: When apache/accumulo#1642 and apache/accumulo#1880 are complete, there will be a bunch of other documentation changes that apply only to the 2.1 release. Is it worth splitting off a separate doc link, or are we better off sticking with that only for major releases?

ctubbsii: > Is it worth splitting off a separate doc link, or are we better off sticking with that only for major releases?

User-facing changes should generally be additive for minor releases, so it would be very redundant to have separate doc areas. I think we can include notes inline where we need to to indicate that something applies to a specific minor release and later, rather than copy/paste docs and maintain them in separate paths, when they are 99% identical.

keith-turner: It may be worthwhile to research practices for annotating since information in documentation and see if there is anything worth adopting.  AFAIK we don't have any standard way to do this in the Accumulo docs.  I have read documentation for other projects where they had a standard way of annotating since information, but I can not remember where I saw that.

ctubbsii: We're using bootstrap 3.3 for the website, currently. Alerts, badges, wells, labels, and even jumbotrons might be suitable. Bootstrap itself uses something called a "callout" in its own documentation, which we could create a custom something like that. https://getbootstrap.com/docs/3.3/components/

PR #237, opened 2020-08-07
Blog post for GH branch renaming

EdColeman:
Short post concerning master branch renaming.

### DIFF
--- a/_posts/blog/2020-08-07-git-branch-name-change.md
+++ b/_posts/blog/2020-08-07-git-branch-name-change.md
@@ -1,0 +1,23 @@
+---
+title: GitHub Branch Renaming
+author: Christopher Tubbs
+---
+
+# Overview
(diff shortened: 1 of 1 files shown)


EdColeman:
My thought was to have something that served as a ready reference.  Not all developers are always active and even if someone sees the email, it may be sometime before they are in a position to update.  Having a basic guide that can be referenced when the are ready seemed to help ease the transition - maybe a blog post is not the right place, maybe the correct place work be under https://accumulo.apache.org/how-to-contribute/ - but that seems "to permanent"

EdColeman:
Put this into draft until INFA makes the needed changes and figure out a better way to communicate this to developers.

ctubbsii:
It's been a few weeks now and there doesn't seem to be any confusion over the branch rename, so, I think we can close this.

PR #245, opened 2020-10-16
Update Gemfile with new versions

milleruntime:
I am not sure if we need to update these or not but they showed up when I ran "bundle update"


milleruntime:
These are recent updates (Sept & Oct):
https://rubygems.org/gems/em-websocket/versions/0.5.2
https://rubygems.org/gems/rouge/versions/3.24.0

arvindshmicrosoft (approved):

ctubbsii (approved):
:+1:  It's good to keep these updated

PR #246, opened 2020-10-16
Improve the LTM wording

milleruntime:
* Simplify some of the wording in the LTM description
* Split up the LTM and non-LTM intents, since some users may
only care about the LTM releases
* Minor spelling corrections
* Clarify the motivation paragraph by making bullets of goals

### DIFF
--- a/contributor/versioning.md
+++ b/contributor/versioning.md
@@ -30,7 +30,7 @@ Goals of the LTM strategy:
 
 Note: the above strategy is a declaration of **intent** only. We use the term
 "Long Term Maintenance" rather than the more familiar "Long Term Support" or
-"Long Term Stable" terms to avoid confusion over the implication of warranties
+"Long Term Stable" terms to avoid any implication of warranties
 from the use of the words "support" or "stable". See the project LICENSE for a full
 disclaimer of warranties. If you have questions, about this, please [contact] us.
 
--- a/contributor/versioning.md
+++ b/contributor/versioning.md
@@ -8,7 +8,7 @@ redirect_from: /versioning
 
 The Apache Accumulo PMC has adopted a Long Term Maintenance (LTM) release strategy.
 
-This strategy entails an intent to:
+### LTM Versions:
 
 1. Periodically release a new LTM `major.minor.0` version (approximately every 2 years),
 2. Maintain the LTM releases with `major.minor.(patch++)` releases until 1 year after the next LTM,
@@ -17,12 +17,12 @@ This strategy entails an intent to:
 This strategy implies that no more than one or two LTM releases will be
 actively maintained at any given time, with one year of overlap.
 
-Non-LTM versions:
+### Non-LTM versions:
 
 1. Release intermediate non-LTM `major.minor.0` versions that are not expected to receive patch/bugfix releases,
 2. Roll patches/bugfixes targeting non-LTM versions into the next `major.minor.0` release
 
-Goals of the LTM strategy:
+### Goals of the LTM strategy:
 - Streamline maintaining multiple versions of Accumulo
 - Encourage development on newer versions
 - Provide greater confidence for users to upgrade and receive updates (within one year)
--- a/contributor/versioning.md
+++ b/contributor/versioning.md
@@ -12,25 +12,27 @@ This strategy entails an intent to:
 
 1. Periodically release a new LTM `major.minor.0` version (approximately every 2 years),
 2. Maintain the LTM releases with `major.minor.(patch++)` releases until 1 year after the next LTM,
-3. Release intermediate non-LTM `major.minor.0` versions that are not expected to receive patch/bugfix releases,
-4. Roll patches/bugfixes targeting non-LTM versions into the next `major.minor.0` release, and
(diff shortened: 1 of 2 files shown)


contributor/versioning.md
```diff
@@ -12,25 +12,27 @@ This strategy entails an intent to:
...
 This strategy implies that no more than one or two LTM releases will be
-actively maintained at any given time, with a one year overlap.
+actively maintained at any given time, with a one-year overlap.
 
-The motivation for this is to streamline the work of maintaining multiple
-versions of Accumulo while not inhibiting development on newer versions, to
-provide greater confidence for risk-averse users to upgrade to versions that
-are expected to be stable and receive updates (with a one year window to do
-so), and help reduce the amount of work done by limiting upgrade paths.
+Non-LTM versions:
```
ctubbsii, line 20: This is a clear Non-LTM section, but there is no corresponding clear LTM section above the earlier list of items. Also, should these headers have `###` section prefixes?

contributor/versioning.md
```diff
@@ -12,25 +12,27 @@ This strategy entails an intent to:
...
 1. Periodically release a new LTM `major.minor.0` version (approximately every 2 years),
 2. Maintain the LTM releases with `major.minor.(patch++)` releases until 1 year after the next LTM,
-3. Release intermediate non-LTM `major.minor.0` versions that are not expected to receive patch/bugfix releases,
-4. Roll patches/bugfixes targeting non-LTM versions into the next `major.minor.0` release, and
-5. Support upgrades from both the immediately preceding LTM and non-LTM versions.
+3. Support upgrades from both the immediately preceding LTM and non-LTM versions
 
 This strategy implies that no more than one or two LTM releases will be
-actively maintained at any given time, with a one year overlap.
+actively maintained at any given time, with a one-year overlap.
```
ctubbsii, line 18: I don't think this is more correct. The hyphen implies this phrase can be treated as a single word, as in an `overlap of type 'one-year'`. I don't think that makes sense grammatically.

If the existing wording isn't clear, we could change it to:
```suggestion
actively maintained at any given time, with one year of overlap.
```

contributor/versioning.md
```diff
@@ -12,25 +12,27 @@ This strategy entails an intent to:
...
 
-The motivation for this is to streamline the work of maintaining multiple
-versions of Accumulo while not inhibiting development on newer versions, to
-provide greater confidence for risk-averse users to upgrade to versions that
-are expected to be stable and receive updates (with a one year window to do
-so), and help reduce the amount of work done by limiting upgrade paths.
+Non-LTM versions:
+
+1. Release intermediate non-LTM `major.minor.0` versions that are not expected to receive patch/bugfix releases,
+2. Roll patches/bugfixes targeting non-LTM versions into the next `major.minor.0` release
```
ctubbsii, line 23: I'm not sure how I feel about them being split up like this. The idea was to describe the entire strategy in one concise list. Spreading it out makes it seem like there are two strategies. It's probably fine separate... I'm just not sure that it's better or not.

contributor/versioning.md
```diff
@@ -12,25 +12,27 @@ This strategy entails an intent to:
...
+- Provide greater confidence for users to upgrade and receive updates (within one year)
+- Help reduce the amount of work done by limiting upgrade paths
 
 Note: the above strategy is a declaration of **intent** only. We use the term
 "Long Term Maintenance" rather than the more familiar "Long Term Support" or
-"Long Term Stable" terms to avoid possible confusion that may arise over
-implication of warrantees from the use of the words "support" or "stable". See
-the project LICENSE for a full disclaimer of warranties. If you have questions,
-about this, please [contact] us.
+"Long Term Stable" terms to avoid confusion over the implication of warranties
```
ctubbsii, line 33: Yours is better. But you might be able to make it even shorter:

```suggestion
"Long Term Stable" terms to avoid any implication of warranties
```

contributor/versioning.md
```diff
@@ -12,25 +12,27 @@ This strategy entails an intent to:
...
+Non-LTM versions:
+
+1. Release intermediate non-LTM `major.minor.0` versions that are not expected to receive patch/bugfix releases,
+2. Roll patches/bugfixes targeting non-LTM versions into the next `major.minor.0` release
+
+Goals of the LTM strategy:
+- Streamline maintaining multiple versions of Accumulo
+- Encourage development on newer versions
+- Provide greater confidence for users to upgrade and receive updates (within one year)
+- Help reduce the amount of work done by limiting upgrade paths
```
ctubbsii, line 29: :+1:  bullet points are way better than prose.

contributor/versioning.md
milleruntime: Yeah I wasn't sure either.  My thinking was that there are probably some managers or non technical folks that do not care about development non LTM releases and I was trying to have a part just for their attention span.

contributor/versioning.md
milleruntime: I updated the sections using ```###``` and replaced the semantic versioning link on the front page to link to this versioning page.  I thought this page is more useful and it contains link to the semantic versioning documentation.

ctubbsii (approved):

PR #256, opened 2021-01-07
initial draft of Apache 2021-01 report

EdColeman:
*** DO NOT MERGE ***  Draft Apache Accumulo January 2021 Report

The Apache Accumulo PMC decided to draft its quarterly board reports on the dev list. Here is a draft of our report which is due Wednesday, Jan 14.  I am hoping that as a PR, collaboration may be easier than emails, however I'm not sure this project is the most appropriate, and hoping for feedback on this as an approach in general and consensus as to the best vehicle going forward.

I will also post to the dev list.

### DIFF
--- a/reports/202101_apache_report.md
+++ b/reports/202101_apache_report.md
@@ -9,9 +9,11 @@ design and is built on top of Apache Hadoop, Zookeeper, and Thrift.
 
 ## Issues:
 
-No change since last report, Oct 2020.  Still waiting for the owner of 
-http://www.accumulodata.com to work with Amazon to reactive the account used for 
-hosting, so it can be pointed to https://accumulo.apache.org or shutdown. Initial emails are at [1].
+The owner of http://www.accumulodata.com no longer has access to the email used to register the 
+domain, and the registrar will not accept communications from alternate accounts.  Currently, the 
+domain is set to expire on 2021-06-28, and without access by the owner, should automatically expire.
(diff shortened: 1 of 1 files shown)


reports/202101_apache_report.md
```diff
@@ -0,0 +1,52 @@
...
+## Description:
+
+The Apache Accumulo sorted, distributed key/value store is a robust, scalable,
+high performance data storage system that features cell-based access control
+and customizable server-side processing. It is based on Google's BigTable
+design and is built on top of Apache Hadoop, Zookeeper, and Thrift.
+
+## Issues:
+
+No change since last report, Oct 2020.  Still waiting for the owner of
```
phrocker, line 12: Maybe add a little context RE why they are unable to at this time? That might help avoid the concern that they are avoiding helping transition ownership.

phrocker (approved):

brianloss (approved):
LGTM. Thanks Ed!

reports/202101_apache_report.md
EdColeman: updated with more current infomation

ctubbsii (changes requested):
I don't think these should be added to the website. They have little value there to website visitors. I don't think it helps collaborate on the contents, either, as it either excludes people who don't follow the GitHub issues, or it requires duplicated effort to post both here and to the mailing list.

PR #257, opened 2021-01-16
Fixes apache/accumulo#1384 documents 2.1.0 GC metadata changes

keith-turner:


brianloss (approved):
Looks like the automated check failed, and it's complaining about the filename. Aside from that the description makes sense to me and as someone who was not familiar with the change, I read this and it's clear to me what action I would need to take.

I resisted the urge to complain about your use of [two spaces after periods](https://www.instructionalsolutions.com/blog/one-space-vs-two-after-period). 😁

ctubbsii (changes requested):
You can just make the file name today's date, and we can update it later when we release. That way, it can be viewed in the staging site. If you include "draft: true" in the front matter, it won't be linked on the main page, and will be clearly marked as draft when viewed with a direct link to it, so even if the draft gets published to the production branch, it won't be a problem.

brianloss (approved):

ctubbsii (approved):

PR #261, opened 2021-01-29
Describes 2.1.0 compaction changes in rel notes

keith-turner:
This change depends on #232.  I created this PR to satisfy apache/accumulo#1617 and ended up documenting all the compactions changes.  This PR describes how user code written against CompactionStrategy may have problems because of CompactionStrategies transitive use of non public API types.   This is described under the compaction section in the rel notes.  I was thinking that alternatively there could be a section in the rel notes about plugins mentioning the compaction strategy changes and changes like apache/accumulo#1891.

### DIFF
--- a/_posts/release/2020-01-19-accumulo-2.1.0.md
+++ b/_posts/release/2020-01-19-accumulo-2.1.0.md
@@ -7,12 +7,50 @@ draft: true
 
 ## Notable Changes
 
+### Compaction Changes
+
+Significant changes were made to how Accumulo compacts files in this release.  See 
+{% dlink administration/compaction %} for details, below are some highlights.
+
+ * Multiple concurrent compactions per tablet on disjoint files is now
+   supported.  Previously only a single compaction could run on a tablet.  This
+   allows tablets that are running long compactions on large files to
+   concurrently compact new smaller files that arrive.
+ * Multiple compaction thread pools per tablet server are now supported.
+   Previously only a single thread pool existed within a tablet server for
+   compactions.  With a single thread pool, if all threads are working on long
+   compactions it can starve quick compactions.  Now compactions with little
+   data can be processed by dedicated thread pools.
+ * Accumulo's default algorithm for selecting files to compact was modified to
+   select the smallest set of files that meet the compaction ratio criteria
+   instead of the largest set.  This change makes tablets more aggressive about
+   reducing their number files while still doing logarithmic compaction work.
+   This change also enables efficiently compacting new small files that arrive
+   during a long running compaction. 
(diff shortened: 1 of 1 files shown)


brianloss (commented):
LGTM. I don't know if this matters, but the date in the file name is 2020-01-19, and I'm guessing the intent was for it to be 2021-01-19. I suspect that will get updated when the release actually happens.

_posts/release/2020-01-19-accumulo-2.1.0.md
```diff
@@ -7,12 +7,50 @@ draft: true
...
+   doing a logarithmic amount of work.
+
+Accumulo has a pluggable user interface called CompactionStrategy, it was
+deprecated in this release and replaced with new interfaces in the public API.
+The javadocs for CompactionStrategy have more detailed information about its
+deprecation.  CompactionStrategy transitively referenced many Accumulo types
+that are not public API.  Therefore it is extremely difficult to know if
+compactions strategies written for one release of Accumulo will work with a
+later release.  The new interfaces that replace CompactionStrategy only use
+public API types making it easier to write code that works across releases.
```
milleruntime, line 46: The last paragraph could be shorter. It could just say:
```suggestion
"The CompactionStrategy interface was deprecated in favor of new interfaces in the public 
API. The javadocs for CompactionStrategy have more detailed information about its deprecation.
```
If the javadoc already includes the deprecation justification, just include a link to the javadoc.

keith-turner: Good idea, I will link to the javadoc.  One irksome thing is I can't test the link ATM

milleruntime: Right, I hear ya.  You could put a link but make the text say something obvious (like REPLACE ME WITH RELEASED JAVADOC) so we know to replace it at release.

keith-turner: I updated the text, it is much shorter.  Actually I can test the link. I tried t locally and it works.  It links to the 2.0.1 version of compactions strategy javadoc ATM.  So I can not test the full experience of linking to the 2.1.0 javadocs, but I can test that the link works.

keith-turner:
> I don't know if this matters, but the date in the file name is 2020-01-19, and I'm guessing the intent was for it to be 2021-01-19. I suspect that will get updated when the release actually happens.

2021 was definitely my intent.   Yeah it probably does not matter, unless it causes someone to make a mistake when setting the correct date.

milleruntime (approved):

PR #264, opened 2021-02-09
Update 2.0.0 release notes for apache/accumulo#1179

ctubbsii:


milleruntime (approved):
This made me think that when we release 2.1 for an LTM release, we should put big 2.0 changes like this in the release notes as well. Perhaps we can just put a link in the 2.1 release notes to the bullets of the 2.0 release notes.

ctubbsii:
> This made me think that when we release 2.1 for an LTM release, we should put big 2.0 changes like this in the release notes as well. Perhaps we can just put a link in the 2.1 release notes to the bullets of the 2.0 release notes.

It is important that users take note of any relevant changes for *every* version since their last version used. We currently have links, forming a chain of versions on the website, to assist users identifying those changes from version to version.

We *could* copy some subset of relevant items (or link to them), on major versions and/or LTM versions. However, I think that would primarily aid users who aren't very diligent about checking release notes, and such users may become dependent on us doing that (and feel it is safe to ignore any changes we might not call out this way), leading to them being more frustrated when they miss something they overlooked that was on a previous release notes page. For diligent users, who are paying attention to the release notes, on the other hand, we'd only be creating clutter, and possibly confusion as the result of duplication of effort.

So, I'd prefer we didn't do that. It's much easier to encourage users to check previous release notes, rather than duplicate effort, barrage diligent users with duplicate information, and deal with the potential downsides of encouraging users to disregard previous release notes under the incorrect assumption that only the current one is irrelevant to them.

PR #266, opened 2021-03-02
External compaction design

dlmarion:


design/external-compaction.md
```diff
@@ -0,0 +1,249 @@
...
+
+  * `tserver.compaction.major.service.<service>.rate.limit`: maximum number of bytes to read or write per second for all compactions in this service.
+
+  * `tserver.compaction.major.service.<service>.planner.opts.*`: other properties used to configure the planner.
+
+
+The DefaultCompactionPlanner implementation is configured to create multiple queues from the configuration where each queue is configured with file size thresholds and the number of threads. For example, the DefaultCompactionPlanner is configured by the following properties:
+
+
+  * `tserver.compaction.major.service.<service>.planner.opts.executors`: a json array where each object in the array has the fields name, maxSize, and numThreads. For example:
```
milleruntime, line 54:
```suggestion
  * `tserver.compaction.major.service.default.planner.opts.executors`: a json array where each object in the array has the fields name, maxSize, and numThreads. For example:
```
The default name should be "default" right?

design/external-compaction.md
```diff
@@ -0,0 +1,249 @@
...
+
+```  
+[
+  {'name':'small','maxSize':'32M','numThreads':2},
+  {'name':'medium','maxSize':'128M','numThreads':2},
+  {'name':'large','numThreads':2}
+]
+```
+
+  * `tserver.compaction.major.service.<service>.planner.opts.maxOpen`: number of files that will be included in a single compaction
```
milleruntime, line 64:
```suggestion
  * `tserver.compaction.major.service.default.planner.opts.maxOpen`: number of files that will be included in a single compaction
```
The default name should be "default" right?

design/external-compaction.md
```diff
@@ -0,0 +1,249 @@
...
+ **/
+void finishedJob(CompactionJob, CompactionStats)
+
+
+/**
+  * Called by compactor process to update the coordinator with the state of the running compaction
+ **/
+void updateCompactionStatus(CompactionJob, State, message, timestamp)
+```
+
```
milleruntime, line 125: I feel like at least one of these methods should include a timeout, I am not sure where that would be determined though.

design/external-compaction.md
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * Atomically updates a tablets files after a compaction finishes. Called by CompactionCoordinator when Compactor finishes compaction
+ */
+void finishedJob(CompactionJob)
+```
+
+## Queue Specification
+
+The CompactionCoordinator is responsible for managing the work queue. For each external compaction queue, the tablet server will maintain an internal in memory priority queue of the tablets loaded on it that have work to do on that external queue. The coordinator polls all tservers to get summary information about their external compaction queues. The coordinator then combines the summary information from all tservers and uses it to determine which tablet server to contact next to get work.  The coordinator does not maintain per tablet information, it only maintains enough information to allow it know which tablet server to contact for a given queue.  The tablet server will then know what specific tablet in that queue needs to compact.  
+
```
milleruntime, line 172: Would it be better to have the tservers push the summary information when it changes? This way we don't want to burden tservers by polling while it is busy.

design/external-compaction.md
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * Called by the coordinator to get summary information about all of a tablets servers external compaction queues.
+ */
+List<TabletServerCompactionQueueSummary> getTabletServerQueueInfo()
+
+
+/**
+ * Called to atomically reserve compaction work for a tablet on a given queue with a priority level >= than what was passed.  Will return null if there is no work meeting criteria.  Tablets will treat these files reserved as part of a running compaction for planning future compactions. Will use CompactorId to check if process running compactions is alive.  Information about the reservation will be persisted by the tablet in the metadata table so that if the tablet loads elsewhere it will not lose track of the external compaction.
+ */
+CompactionJob reserveJob(String queueName, long priority, CompactorId)
```
milleruntime, line 160: Would the priority be system generated (like age + size)?

design/external-compaction.md
```diff
@@ -0,0 +1,249 @@
...
+ 1. Tablet T1 has files F1,F2,F3,F4
+ 1. Compaction of files F1,F2,F3 is queued on internal executor IE1 for Tablet T1
+ 1. New files F5,F6,F7,F8,F9 arrive for tablet T1
+ 1. The compaction of F1,F2,F3 on IE1 is canceled and a new compaction of F1,F2,F3,F4,F5,F6,F7,F8,F9 is queued on external Executor EE2 for Tablet T1
+
+Scenario 3
+1. Tablet T1 has files F1,F2,F3
+2. Compaction of files F1,F1,F3 is queued on external executor EE2 for T1
+3. The table containing tablet T1 is deleted
+4. A compactor tries to compact F1,F2,F3 for tablet T1 and hopefully fails cleanly
```
milleruntime, line 224: I think it would be better to call cancel on any compactions if a table is deleted.

design/external-compaction.md
```diff
@@ -0,0 +1,249 @@
...
+## Scenarios
+
+Scenario 1
+
+ 1. Tablet T1 has files F1,F4,F5,F6
+ 1. The compaction of files F4,F5,F6 is queued on external executor EE2 for Tablet T1
+ 1. A new file F7 arrives for T1
+ 1. The compaction of F4,F5,F6 on EE2 is canceled and new a compaction of F4,F5,F6,F7 is queued on EE2 for Tablet T1
+ 1. A new file F8 arrives for T1
+ 1. The compaction of F4,F5,F6,F7 on EE2 is canceled and new a compaction of F4,F5,F6,F7,F8 is queued on EE2 for Tablet T1
```
milleruntime, line 211: Maybe it would be good to have different configurable options for handling this scenario. Option 1 would be to cancel any compactions on the arrival of new files. Option 2 would be to allow the compaction to finish (not cancel on arrival). This would allow users to tune their cluster for faster scans or cleaning old data. This might also be complicated by whether the compaction is a USER or system type.  For example, we may want to allow user compactions to finish vs cancelling a system one.

design/external-compaction.md
```diff
@@ -0,0 +1,249 @@
...
+6. The compactor is able to commit the compaction of F1,F2,F3 for T1
+
+
+Scenario 6
+1. Compaction of files F1,F1,F3 is queued on external executor EE2 for T1
+2. Compaction of files F4,F5,F6 is queued on external executor EE2 for T2
+3. A compactor start compacting F1,F2,F3 for T1
+4. The table hosting tablets T1 and T2 is taken offline
+5. The compactor working on F1,F2,F3 for T1 finishes and cleanly fails to commit the compaction because the table is offline
+6. The compaction for F4,F5,F6 is taken out of the queue, but cleanly fails to start because the table is offline.
```
milleruntime, line 249: I know this design is strictly for online compactions but we could make offline compactions configurable. Maybe even have a separate queue for handling offline compactions.

milleruntime (commented):
Looks good overall. I provided some thoughts for consideration.

design/external-compaction.md
milleruntime: Also, I am not sure if we want to cancel compactions with the arrival of new data. I don't think we currently do that at all.

design/external-compaction.md
dlmarion: If you can modify the behavior of the default compaction planner by using these properties, then yes.

design/external-compaction.md
keith-turner: > Would it be better to have the tservers push the summary information when it changes?

I am uncertain which is better, but I am leaning a bit towards pushing.  One advantage to tservers pushing is that tservers w/ nothing in their external queues would not push anything.  With polling, it would keep polling tservers that have no tablets in their external queues.

Was thinking polling is better in the case when the coordinator starts up.  It can do something like the following.

 * poll all tservers
 * start accepting request for work from compactors

With push, I suppose it could do something like the following

 * wait X minutes for tservers to report external compaction summary info
 * start accepting request for work from compactors

So need to figure out the coordinator startup behavior w/ push.

design/external-compaction.md
keith-turner: The default planner currenty uses the compaction type and #of files to construct a priority.   The following are links to the code that currently creates the priority.

https://github.com/apache/accumulo/blob/b12d1103d788473168616f28ae87a65bb76aa880/core/src/main/java/org/apache/accumulo/core/spi/compaction/DefaultCompactionPlanner.java#L236

https://github.com/apache/accumulo/blob/b12d1103d788473168616f28ae87a65bb76aa880/core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionJobPrioritizer.java#L33

@dlmarion  and I were talking and concluded that long is probably too high cardinality for priority.  Maybe a 16 bit or 8 bit integer would be better.   Then maybe the prio could be type:log2(numFiles)

design/external-compaction.md
keith-turner: The wording needs to be updated.  The compaction is not yet running and cancel just means its canceled in the queue.  The current code never cancels a running compaction when the plan changes, only queued compactions.

design/external-compaction.md
milleruntime: What about only polling at startup so compactors can get started right away? I suppose a hybrid approach would then require 2 different RPC endpoints though. Maybe that's not such a big deal?

keith-turner: The hybrid approach may be best, I was hoping to think of something simpler.

design/external-compaction.md
milleruntime: Ah OK. So something more like "EE2 is removed from the queue"

design/external-compaction.md
keith-turner: If the tserver push frequency is known, like tservers push at least every 30 secs.  Then the coordinator could wait 2x that I think and it would be close to optimal.  If the coordinator misses a few updates from tservers before handing out work, then it may start some lower priority compactions running before higher prio ones.  That is probably ok.

design/external-compaction.md
keith-turner: "Removed from the queue" would be better working.  In the code it actually just cancels it, possibly leaving it in the queue for efficiency.  Something canceled in the queue would never run.  However those are just implementation details, its doing lazy removal.

Below is where only queued compactions are canceled

https://github.com/apache/accumulo/blob/b12d1103d788473168616f28ae87a65bb76aa880/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionService.java#L185

Below is where a task is atomically transitioned to cancel, but only if its currently in the queued state.

https://github.com/apache/accumulo/blob/b12d1103d788473168616f28ae87a65bb76aa880/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionExecutor.java#L112

Below is where a task atomically transitions from queued state to running.  The task in the canceled state, then the transition will fail.

https://github.com/apache/accumulo/blob/b12d1103d788473168616f28ae87a65bb76aa880/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionExecutor.java#L88

PR #268, opened 2021-03-31
Add prominent link to public API docs

keith-turner:


dlmarion (approved):

keith-turner:
I thought there used to be a prominent link to this, however I can not find it if there is.

milleruntime:
Closes https://github.com/apache/accumulo-website/issues/207

milleruntime:
There is also a link on the main page under "Stable" feature to "client api" which we may want to change to this as well.

keith-turner:
> There is also a link on the main page under "Stable" feature to "client api" which we may want to change to this as well.

That seems to link to Javadoc that only includes the public API.  If so maybe the /api page should link to this javadoc and then "client api" link under "Stable" could link to /api

PR #281, opened 2021-05-05
Added 'Akshit Mangotra' in people.md

Akshit6828:
https://github.com/apache/accumulo/pull/2080

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@ the License.  You may obtain a copy of the License at
 
     http://www.apache.org/licenses/LICENSE-2.0
 
-Unless required by applicable law or agreed to in writing, software
+Unless required by applicable law or agreed to in writing, softwares
 distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -8,7 +8,7 @@ the License.  You may obtain a copy of the License at
 
     http://www.apache.org/licenses/LICENSE-2.0
 
-Unless required by applicable law or agreed to in writing, software
+Unless required by applicable law or agreed to in writing, softwares
```
EdColeman, line 11: Please revert this change.

1) The plural form of software is software - so the original is correct.  
2) This is boilerplate from the ASF, if it would need to be changed it should be done at the ASF level.

EdColeman (changes requested):

CONTRIBUTING.md
Akshit6828: Oops! Sorry @EdColeman, sorry my bad. I didn't do that intentionally as I was just going through the readme in vs code and it might have pressed some key. I am sorry for this change. I'll revert is ASAP.

PR #285, opened 2021-06-15
Update tools doc for CheckForMetadataProblems

Manno15:
From Jira ticket: https://issues.apache.org/jira/browse/ACCUMULO-2845. A small change in output. Now tools doc lines up with the actual output.

### DIFF
--- a/_docs-2/troubleshooting/tools.md
+++ b/_docs-2/troubleshooting/tools.md
@@ -102,8 +102,13 @@ every tablet matches, and the start and stop for the table is empty:
 
     $ accumulo org.apache.accumulo.server.util.CheckForMetadataProblems -u root --password
     Enter the connection password:
+    Checking table: accumulo.root
     All is well for table !0
+    No problems found
+    Checking table: accumulo.metadata
+    All is well for table +rep
     All is well for table 1
+    No problems found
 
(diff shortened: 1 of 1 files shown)


ctubbsii (approved):

_docs-2/troubleshooting/tools.md
```diff
@@ -102,8 +102,13 @@ every tablet matches, and the start and stop for the table is empty:
 
     $ accumulo org.apache.accumulo.server.util.CheckForMetadataProblems -u root --password
     Enter the connection password:
+    Checking table: accumulo.root
     All is well for table !0
```
EdColeman, line 106: Might be out of scope for this PR - but it might be nice if the tool also used names so that it was like accumlo.metadata (!0), accumulo.root (+r)  or something consistently.

This output seems slightly misleading - it reports checking root but says all is well for the metadata (!0) 

The same for trace (table 1)

But this would be more of a tool improvement rather than update docs to match what's displayed right now.

Manno15: Good idea. Feel free to create an issue for in accumulo and I will take a look at it. Possible good first issue as well but I do not know off the top of my head.

PR #287, opened 2021-07-07
Update tools.md for new CheckForMetadataProblems output

foster33:
Since there were changes made to CheckForMetadataProblems' output [here](https://github.com/apache/accumulo/pull/2180), there needs to be an update to the tools doc that shows the changes. 

This PR updates the tools doc to show the new output.


Manno15 (approved):

ctubbsii (commented):
Is the new output both indented with spaces *and* prefixed with `...`? I thought only the latter. If it's also indented with spaces, we should only use one.

foster33:
> Is the new output both indented with spaces _and_ prefixed with `...`? I thought only the latter. If it's also indented with spaces, we should only use one.

It is only `...`  . Good catch. I am not sure why I included the spaces. I will remove the spaces.

ctubbsii (approved):

ctubbsii:
Thanks @foster33 !

PR #303, opened 2021-12-13
Update release notes for new monitor page and utility

milleruntime:


_posts/release/2020-01-19-accumulo-2.1.0.md
```diff
@@ -213,6 +213,12 @@ TABLE: accumulo.metadata
 2    default_tablet  0     0     0         0         HOSTED     CURRENT:ip-10-113-12-25:9997   !0    ~                    +INF
 ```
 
+### New Utility for generating splits
+
+A new command line utility was created to generate split points from 1 or more rfiles. One or more HDFS directories can
+be given as well. The utility will iterate over all the files provided and determine the proper split points
+based on either the size or number given. It uses Apache Datasketches to get the split points from the data. {% ghi 2361 %} {% ghi 2368 %}
+
```
EdColeman, line 221: Could this be expanded?  

- Are the split points generated evenly across the row ids seen, or are they generated so that when applied to the the files the number of splits in each file are approximately equal?
- How are duplicate row ids counted? As one, or each duplicate contributing to the count?
- When specifying size, does the size of the values matter?

And I think the datasketches is an implementation detail that is either unnecessary - or could be expanded as to why that it is important.

milleruntime: It could but I only wanted to give a short, high level description for the release notes. I think some of your questions are details that would be better suited for the javadoc or description of the utility.

Mentioning Apache datasketches is really just a shout out for use of other Apache tools.

EdColeman (approved):

PR #304, opened 2021-12-20
Update tracing documentation for 2.1

dlmarion:


_docs-2/troubleshooting/tracing.md
```diff
@@ -16,328 +16,44 @@ enabled follows all the requests made on behalf of the user throughout
...
+and invokes the Accumulo client operation in a Span. Client application
+developers can use the OpenTelemetry [documentation] to instrument the
+application. To collect traces in the client, Accumulo needs the same two
+items mentioned previously: `general.opentelemetry.enabled` in the client
+application and the `io.opentelemetry.api.GlobalOpenTelemetry.globalOpenTelemetry`
+member variable must be set.
 
 ### Tracing from the Shell
 You can enable tracing for operations run from the shell by using the
 `trace on` and `trace off` commands.
```
EdColeman, line 53: Is this still true?

dlmarion: Yes. See [this](https://github.com/apache/accumulo/blob/main/core/src/main/java/org/apache/accumulo/core/trace/TraceUtil.java#L56)

dlmarion: Ok, so that might not be correct for `clients`. I will update that statement.

DomGarguilo (approved):

PR #323, opened 2022-06-14
Add sections for pluggable block caches and scan execution hints that control caching

dlmarion:


_docs-2/administration/caching.md
```diff
@@ -18,6 +18,10 @@ for tables where read performance is critical.
 
 ## Configuration
 
+The {% plink tserver.cache.manager.class %} property controls which block cache implementation is used within the tablet server. Users
+can supply their own implementation and set custom configuration properties to control it's behavior (see org.apache.accumulo.core.spi.cache.BlockCacheManager$Configuration.java).
+An example off-heap block cache implementation and configuration can be found at: https://github.com/keith-turner/accumulo-ohc.
+
```
ctubbsii, line 24: An example would serve great as a block post, but I don't think it's particularly helpful as a mere link to one of many possible external implementations.

ctubbsii:
> I'm wondering if we need to have @keith-turner contribute his https://github.com/keith-turner/accumulo-ohc repo under the Apache umbrella.

No, we don't *need* to. The point of creating SPIs is that anybody can create and publish their own implementation, without the need to ship all the implementations in Accumulo itself. SPIs are integration points that can be used to expand the ecosystem of related projects, use-cases, and users. The whole point is that they don't need to be *here* to be part of that ecosystem. They can be maintained by users and live elsewhere.

If the author of an implementation wishes to donate their implementation to the ASF, they can do so... they can even contribute it directly to the Accumulo PMC. But the decision to accept it would be something the PMC would need to deliberate on. Personally, I think I would need to find it very compelling to accept the added responsibility for it... because I don't see a lot of extra value to users for it to live *here* vs. living *there*, but there is certainly added responsibility on our part if we accept it.

_docs-2/administration/caching.md
dlmarion: > An example would serve great as a block post

Did you mean adding a code block to caching.md, or a blog post?

ctubbsii: I meant blog post. Sorry for the typo.

ctubbsii (approved):

PR #330, opened 2022-07-20
Add JShell version of the Accumulo Tour to website

jmark99:
Added a version of the Accumulo Tour that makes use of the JShell feature of Accumulo 2.1.x. The user will have to option to go through the tour using the former version with the MiniAccumuloCluster or the newer JShell feature.

Also added a new page to the tour illustrating iterators.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -46,6 +46,13 @@ defaults:
       permalink: "/tour/:basename/"
   -
     scope:
+      path: "tour-jshell"
+      type: "pages"
+    values:
+      layout: "tour-jshell"
+      permalink: "/tour-jshell/:basename/"
+  -
+    scope:
       path: "_posts/blog"
       type: "posts"
     values:
--- a/tour-jshell/basic-read-write2.md
+++ b/tour-jshell/basic-read-write2.md
@@ -1,0 +1,80 @@
+---
+title: Writing and Reading
+---
+
+Accumulo is a big key/value store.  Writing data to Accumulo is flexible and fast.  Like any 
+database, Accumulo stores data in tables and rows.  Each row in an Accumulo table can hold many 
+key/value pairs. 
+
+Our next exercise shows how to write and read from a table.
+
+Let's create a table called "GothamPD".
+
+At the JShell prompt, enter the following:
+```commandline
+jshell> client.tableOperations().create("GothamPD");
+```
+
+Accumulo uses Mutation objects to hold all changes to a row in a table. Each row has a unique row
+ID. 
+
+```commandline
+jshell> Mutation mutation1 = new Mutation("id0001");
+mutation1 ==> org.apache.accumulo.core.data.Mutation@1
+```
+
+Create key/value pairs for Batman.  Put them in the "hero" family.
+```commandline
+jshell> mutation1.put("hero","alias", "Batman");
+jshell> mutation1.put("hero","name", "Bruce Wayne");
+jshell> mutation1.put("hero","wearsCape?", "true");
+```
+
+Create a BatchWriter to the GothamPD table and add your mutation to it. Try-with-resources will 
+close for us.
+
+```commandline
+jshell> try (BatchWriter writer = client.createBatchWriter("GothamPD")) {
+  ...>    writer.addMutation(mutation1);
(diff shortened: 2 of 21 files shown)


tour-jshell/ranges-splits2.md
```diff
@@ -0,0 +1,44 @@
+---
+title: Ranges and Splits
+---
+
+A [Range] is a specified group of `Key`s. There are many ways to create a `Range`.  Here are a few examples:
+
```
milleruntime, line 6: It looks like some of the pages are the same as the Mini tour? Might be able to reuse the pages between the tours.

milleruntime (approved):

tour-jshell/index.md
```diff
@@ -0,0 +1,26 @@
...
+If you have any questions or suggestions while
+going through the tour, please email our [mailing list][mlist] or [create an issue][issue].
+
+{% for p in tour_pages %}
+  {% assign doc_url = p | prepend: '/tour-jshell/' | append: '/' %}
+  {% assign link_to_page = site.pages | where:'url',doc_url | first %}
+  1. [{{ link_to_page.title }}]({{ doc_url }})
+{% endfor %}
+
+[mlist]: /contact-us/#mailing-lists
```
DomGarguilo, line 25:
```suggestion
[mlist]: https://accumulo.apache.org/contact-us/contact-us/#mailing-lists
```
My IDE complains about this link unless its the whole URL.

tour-jshell/ranges-splits2.md
jmark99: Although a couple of pages contain very similar content, there were several formatting changes that cause them to differ. It may be possible to replace the older tour page with the updated jshell tour page. Not sure how difficult that would be. Can look into it later.

tour-jshell/basic-read-write2.md
```diff
@@ -0,0 +1,80 @@
...
+
+### But wait... I thought Accumulo was all about Security?
+
+Spoiler Alert: It is!  Did you notice the `Authorizations.EMPTY` we passed in when creating a
+[Scanner]?  The data we created in this first lesson was not secured with Authorizations so the 
+Scanner didn't require any Authorizations to read it.  More to come later in the [Authorizations][auths] 
+lesson!
+
+[dmodel]: /tour-jshell/data-model2
+[auths]: /tour-jshell/authorizations2
```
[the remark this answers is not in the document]
jmark99: I tried the changes and it would not work on my local web server. Compared these links to similar existing tour pages and then format is the same. Will keep an eye on this once it is published and update then if necessary.

tour-jshell/index.md
jmark99: As above, this link matches the format of the existing tour page. Again, will monitor once published and change then if necessary.

tour-jshell/basic-read-write2.md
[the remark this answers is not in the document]
DomGarguilo: Sounds good :+1:

_config.yml
```diff
@@ -44,6 +44,13 @@ defaults:
     values:
       layout: "tour"
       permalink: "/tour/:basename/"
+  -
+    scope:
+      path: "tour-jshell"
+      type: "pages"
+    values:
+      layout: "tour-jshell"
+      permalink: "/tour-jshell/:basename/"
```
ctubbsii, line 53: This should replace the existing tour. Not sit adjacent to it.

jmark99: If we replace the existing tour then users using any 1.x or 2.0.x version would not be able to take advantage of the tour. I would think at least for a while it would be best to keep both accessible.

ctubbsii: JShell can be used with the older version. We just don't have the convenience tools to automatically set up the class path. The jshell-based tour would still be applicable for those versions, but setting up the classpath could be left as an exercise to the user.

Also, the tour targets new users. New users are not typically starting off with old versions.

I would also call the tour "Interactive" rather than "JShell". JShell is the just the tool, but the point of the tour using JShell is to be more interactive, so that's the characteristic I would emphasize in any menus/navigation/page titles.

PR #334, opened 2022-07-28
Replace older Tour with newer JShell Tour

jmark99:
This update completely removes the older Accumulo Tour code with the newer interactive JShell version of the Tour.


tour/batch-scanner-code.md
```diff
@@ -4,44 +4,52 @@ title: Batch Scanner Code
...
+   ...>   batchScanner.fetchColumn(new Text("villain"), new Text("yearsOfService"));
+   ...> 
+   ...>   // Calculate average years of service
+   ...>   Long totalYears = 0L;
+   ...>   Long entriesRead = 0L;
+   ...>   for (Map.Entry<Key, Value> entry : batchScanner) {
+   ...>     totalYears += Long.valueOf(entry.getValue().toString());
+   ...>     entriesRead++;
+   ...>   }
+   ...>   System.out.println("The average years of service of " + entriesRead + " villains is " + totalYears / entriesRead);
```
DomGarguilo, line 49: It could be cool to show off how BatchScanner.stream could be used here or somewhere else.

For example, could be used to find the average like this:
```
Double average = scanner.stream().map(Map.Entry::getValue).map(Value::toString).mapToLong(Long::valueOf).average().getAsDouble();
```

DomGarguilo (approved):
Everything looks good to me

tour/batch-scanner-code.md
jmark99: @DomGarguilo I updated the batchScanner to make use of streams as suggested. Thanks for the suggestion.

PR #341, opened 2022-09-23
Add setting multiple properties using java api

EdColeman:
Documents ability to modify properties in a single operation provided in https://github.com/apache/accumulo/pull/2799

### DIFF
--- a/_docs-2/configuration/overview.md
+++ b/_docs-2/configuration/overview.md
@@ -59,6 +59,17 @@ They can also be set using {% jlink org.apache.accumulo.core.client.admin.Instan
 client.instanceOperations().setProperty("table.durability", "flush");
 ```
 
+The java api also supports adding, modifying and removing multiple properties in a single operation:
+
+```java
+client.instanceOperations().modifyProperties(properties -> {
+  properties.remove(Property.TABLE_FILE_MAX.getKey());
+  properties.put("table.bloom.enabled", "true");
+  properties.put("table.bloom.error.rate", "0.75");
+  properties.put("table.bloom.size", "128000");
+});
+
+```
 ### Namespace
 
 Namespace configuration refers to [table.* properties] set for a certain table namespace (i.e group of tables). These settings are stored in ZooKeeper. Namespace configuration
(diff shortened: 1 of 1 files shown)


_docs-2/configuration/overview.md
```diff
@@ -59,6 +59,17 @@ They can also be set using {% jlink org.apache.accumulo.core.client.admin.Instan
 client.instanceOperations().setProperty("table.durability", "flush");
 ```
 
+The java api also supports adding, modifying and removing multiple properties in a single operation:
+
+```java
+client.instanceOperations().modifyProperties(properties -> {
+  properties.remove(Property.TABLE_FILE_MAX.getKey());
```
dlmarion, line 66: Can you set table properties at the instance level?

EdColeman: Yes - with set on an instance the become "system" level properties.

Before:

```
root@uno> config -f bloom
-----------+--------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------
SCOPE      | NAME                                 | VALUE
-----------+--------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------
default    | table.bloom.enabled ................ | false
default    | table.bloom.error.rate ............. | 0.5%
default    | table.bloom.hash.type .............. | murmur
default    | table.bloom.key.functor ............ | org.apache.accumulo.core.file.keyfunctor.RowFunctor
default    | table.bloom.load.threshold ......... | 1
default    | table.bloom.size ................... | 1048576
default    | tserver.bloom.load.concurrent.max .. | 4
-----------+--------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------

```

After:

```
jshell> import org.apache.accumulo.core.conf.Property;

jshell> client.instanceOperations().modifyProperties(properties -> {
   ...>   properties.remove(Property.TABLE_FILE_MAX.getKey());
   ...>   properties.put("table.bloom.enabled", "true");
   ...>   properties.put("table.bloom.error.rate", "0.75");
   ...>   properties.put("table.bloom.size", "128000");
   ...> });

root@uno> config -f bloom
-----------+--------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------
SCOPE      | NAME                                 | VALUE
-----------+--------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------
default    | table.bloom.enabled ................ | false
system     |    @override ....................... | true
default    | table.bloom.error.rate ............. | 0.5%
system     |    @override ....................... | 0.75
default    | table.bloom.hash.type .............. | murmur
default    | table.bloom.key.functor ............ | org.apache.accumulo.core.file.keyfunctor.RowFunctor
default    | table.bloom.load.threshold ......... | 1
default    | table.bloom.size ................... | 1048576
system     |    @override ....................... | 128000
default    | tserver.bloom.load.concurrent.max .. | 4
-----------+--------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------

```

dlmarion (approved):

_docs-2/configuration/overview.md
ctubbsii, line 66: `org.apache.accumulo.core.conf.Property` is not public API, and should not be used in documentation as an example. It should be replaced with the actual String key for the property, which is the user-facing "public API" for configuration.

ctubbsii: Fixed in 4dbd6b5f95a406d790fcbbd5182ed9fbcdef3f75

PR #342, opened 2022-09-26
Per table crypto documentation updates

cshannon:
Here are in progress updated for the documentation for per table crypto. Creating a draft pull request for now so it can be reviewed for errors.


cshannon:
@milleruntime and @dlmarion - I had some time tonight so I started working on the documentation updates for crypto. I still need to do some more polishing and fill in some more details but feel free to take a look and see what you think so far. The main doc changes so far should be good enough at this point to set up instance wide or per table crypto now with the new properties.

cshannon:
I forgot to add I may have time to work on this PR a little bit more Wednesday, otherwise mostly I plan to finish it up on Friday morning this week so if you have any feedback before then that would be helpful.

_docs-2/configuration/server-properties.md
```diff
@@ -25,6 +25,7 @@ Below are properties set in `accumulo.properties` or the Accumulo shell that con
...
 | <a name="general_context_class_loader_factory" class="prop"></a> general.context.class.loader.factory | Name of classloader factory to be used to create classloaders for named contexts, such as per-table contexts set by `table.class.loader.context`.<br>**type:** CLASSNAME, **zk mutable:** no, **default value:** empty |
 | <a name="general_custom_prefix" class="prop"></a> **general.custom.*** | Prefix to be used for user defined system-wide properties. This may be particularly useful for system-wide configuration for various user-implementations of pluggable Accumulo features, such as the balancer or volume chooser. |
+| <a name="general_custom_crypto_service" class="prop"></a> general.custom.crypto.service | **Experimental.** The class which executes on-disk file encryption. The default does nothing. To enable encryption, replace this classname with an implementation of theorg.apache.accumulo.core.spi.crypto.CryptoService interface.<br>**type:** CLASSNAME, **zk mutable:** no, **default value:** {% jlink -f org.apache.accumulo.core.spi.crypto.NoCryptoService %} |
```
dlmarion, line 28:
```suggestion
| <a name="general_custom_crypto_service" class="prop"></a> general.custom.crypto.service | **Experimental.** The class which executes on-disk file encryption. The default does nothing. To enable encryption, replace this classname with an implementation of the org.apache.accumulo.core.spi.crypto.CryptoService interface.<br>**type:** CLASSNAME, **zk mutable:** no, **default value:** {% jlink -f org.apache.accumulo.core.spi.crypto.NoCryptoService %} |
```

_docs-2/security/on-disk-encryption.md
```diff
@@ -5,23 +5,57 @@ order: 5
 ---
 
 For an additional layer of security, Accumulo can encrypt files stored on disk.  On Disk encryption was reworked 
-for 2.0, making it easier to configure and more secure.  The files that can be encrypted include: [RFiles][design] and Write Ahead Logs (WALs). NOTE: This feature is considered experimental. For more information, see the [notes below](#things-to-keep-in-mind).
+for 2.0, making it easier to configure and more secure.  Starting with 2.1 On Disk Encryption can now be configured
+per table as well as for the entire instance (all tables). The files that can be encrypted include: [RFiles][design] and Write Ahead 
+Logs (WALs). NOTE: This feature is considered experimental. For more information, see the [notes below](#things-to-keep-in-mind).
```
dlmarion, line 11: I think we should drop another note here that points people to _docs-2/administration/upgrading.md. Then, in upgrading.md we should add a warning that upgrading an encrypted instance to 2.1.0 is not currently supported.

_docs-2/security/on-disk-encryption.md
```diff
@@ -57,6 +91,13 @@ required to perform decryption. The FileDecrypter only has one method:
 ```
 For more help getting started see {% jlink org.apache.accumulo.core.security.crypto.impl.AESCryptoService %}.
 
+## Disabling Crypto
+
+Crypto can be disabled by setting the property `general.custom.crypto.enabled` to false.
```
dlmarion, line 96: Does this disable for the entire instance? Is there a way to disable for a table when using the PerTableCryptoServiceFactory?

cshannon: Good question, this was something quick that I used just to test if I could disable it last week and have things still work and this is a section that needs to be elaborated on and isn't quite finished. I will take a look tomorrow and see where it gets applied as off the top of my head I don't remember if this only works for instance wide crypto and if there is a different property for per table to disable.

milleruntime: See my comment here: https://github.com/apache/accumulo/issues/2930#issuecomment-1249615992

_docs-2/security/on-disk-encryption.md
```diff
@@ -5,23 +5,57 @@ order: 5
...
+To encrypt per table, the per table crypto service factory can be used, `PerTableCryptoServiceFactory`. This factory
+will load a crypto service configured by table. 
+```
+instance.crypto.opts.factory=org.apache.accumulo.core.spi.crypto.PerTableCryptoServiceFactory
+```
+
+The PerTableCryptoServiceFactory requires configuring a crypto service to load for the table RFiles and this can be done by adding the
+{% plink table.crypto.opts.service %} property to a table. Example in the accumulo shell:
+```
+createtable table1 -prop table.crypto.opts.service=org.apache.accumulo.core.spi.crypto.AESCryptoService
```
dlmarion, line 45: what happens if you set `table.crypto.opts.service=org.apache.accumulo.core.spi.crypto.AESCryptoService` in accumulo.properties? Are all tables encrypted then?

milleruntime: This would be the same as any table properties set at the site configuration (accumulo.properites), all tables would get this property. You would still need to set the Factory for it to actually work though.

cshannon: I haven't tried that but I would assume it would do nothing and be ignored. That property should only apply at the per table context but @milleruntime can confirm or I can test it when I get a chance to verify.

cshannon: Ah @milleruntime  just beat me to it, so I guess it would apply to all tables.

dlmarion: thx. I'm working on integrating this into the terraform configuration. I did not put this property into accumulo.properties.

_docs-2/security/on-disk-encryption.md
```diff
@@ -5,23 +5,57 @@ order: 5
 ---
 
 For an additional layer of security, Accumulo can encrypt files stored on disk.  On Disk encryption was reworked 
-for 2.0, making it easier to configure and more secure.  The files that can be encrypted include: [RFiles][design] and Write Ahead Logs (WALs). NOTE: This feature is considered experimental. For more information, see the [notes below](#things-to-keep-in-mind).
+for 2.0, making it easier to configure and more secure.  Starting with 2.1 On Disk Encryption can now be configured
```
DomGarguilo, line 8:
```suggestion
for 2.0, making it easier to configure and more secure.  Starting with 2.1, On Disk Encryption can now be configured
```

_docs-2/security/on-disk-encryption.md
```diff
@@ -5,23 +5,57 @@ order: 5
...
+Logs (WALs). NOTE: This feature is considered experimental. For more information, see the [notes below](#things-to-keep-in-mind).
 
 ## Configuration
 
-To encrypt all tables on disk, encryption must be enabled before an Accumulo instance is initialized.  If on disk 
-encryption is enabled on an existing cluster, only files created after it is enabled will be encrypted 
-(root and metadata tables will not be encrypted in this case) and existing data won't be encrypted until compaction.  To configure on disk encryption, add the 
-{% plink instance.crypto.service %} property to your `accumulo.properties` file.  The value of this property is the
+To encrypt tables on disk, encryption must be enabled before an Accumulo instance is initialized. This is
+done by configuring a crypto service factory. If on disk encryption is enabled on an existing cluster, only files
```
DomGarguilo, line 15:
```suggestion
done by configuring a crypto service factory. If on-disk encryption is enabled on an existing cluster, only files
```

DomGarguilo: looks like there are multiple places that could make this change

_docs-2/security/on-disk-encryption.md
```diff
@@ -5,23 +5,57 @@ order: 5
...
+
+### Per Table Encryption
+
+To encrypt per table, the per table crypto service factory can be used, `PerTableCryptoServiceFactory`. This factory
+will load a crypto service configured by table. 
+```
+instance.crypto.opts.factory=org.apache.accumulo.core.spi.crypto.PerTableCryptoServiceFactory
+```
+
+The PerTableCryptoServiceFactory requires configuring a crypto service to load for the table RFiles and this can be done by adding the
```
DomGarguilo, line 42:
```suggestion
The `PerTableCryptoServiceFactory` requires configuring a crypto service to load for the table RFiles and this can be done by adding the
```

_docs-2/security/on-disk-encryption.md
```diff
@@ -5,23 +5,57 @@ order: 5
...
+{% plink table.crypto.opts.service %} property to a table. Example in the accumulo shell:
+```
+createtable table1 -prop table.crypto.opts.service=org.apache.accumulo.core.spi.crypto.AESCryptoService
+```
+The PerTableCryptoServiceFactory also requires configuring a recovery and WAL crypto service by adding the following
```
DomGarguilo, line 47:
```suggestion
The `PerTableCryptoServiceFactory` also requires configuring a recovery and WAL crypto service by adding the following
```

_docs-2/configuration/server-properties.md
milleruntime: This file is generated so the edit will have to go in Property.java. 
`<!-- WARNING: Do not edit this file. It is a generated file that is copied from Accumulo build (from core/target/generated-docs) -->`

milleruntime (commented):
This file is generated in the main Accumulo build (see the warning at the top of the file) `_docs-2/configuration/server-properties.md`. I wasn't sure if you copied it or manually edited but it should just be copied from the latest in `core/target/generated-docs/`. If we see something wrong with the properties, then we can edit it in Accumulo main. I think they are all just pulled from Property.java.

_docs-2/security/on-disk-encryption.md
```diff
@@ -5,23 +5,57 @@ order: 5
...
 ## Configuration
 
-To encrypt all tables on disk, encryption must be enabled before an Accumulo instance is initialized.  If on disk 
-encryption is enabled on an existing cluster, only files created after it is enabled will be encrypted 
-(root and metadata tables will not be encrypted in this case) and existing data won't be encrypted until compaction.  To configure on disk encryption, add the 
-{% plink instance.crypto.service %} property to your `accumulo.properties` file.  The value of this property is the
+To encrypt tables on disk, encryption must be enabled before an Accumulo instance is initialized. This is
+done by configuring a crypto service factory. If on disk encryption is enabled on an existing cluster, only files
+created after it is enabled will be encrypted (root and metadata tables will not be encrypted in this case) and
+existing data won't be encrypted until compaction.
```
milleruntime, line 17: The root and metadata tables should be encrypted once they are compacted, assuming the user configured it this way.
```suggestion
created after it is enabled will be encrypted and existing data won't be encrypted until compaction. 
```

_docs-2/security/on-disk-encryption.md
```diff
@@ -5,23 +5,57 @@ order: 5
...
+### Encrypting All Tables
+
+To encrypt all tables, the generic crypto service factory can be used, `GenericCryptoServiceFactory`. This factory
+is useful for general purpose on disk encryption with no table context.
+```
+instance.crypto.opts.factory=org.apache.accumulo.core.spi.crypto.GenericCryptoServiceFactory
+```
+
+The GenericCryptoServiceFactory requires configuring a crypto service to load and this can be done by adding the
+{% plink general.custom.crypto.service %} property to your `accumulo.properties` file.  The value of this property is the
```
milleruntime, line 28: You could just tell users to set the property, since there are multiple ways to do this.
```suggestion
The GenericCryptoServiceFactory requires configuring a crypto service to load and this can be done by setting the
{% plink general.custom.crypto.service %} property.  The value of this property is the
```

dlmarion (changes requested):

_docs-2/configuration/server-properties.md
cshannon: @milleruntime - I assume I don't need to bother copying now and that we will just copy the generated file later when we are ready and things are finished from the final release?

milleruntime: Yeah. You can just have this PR modify the on-disk-encryption section.

cshannon: Sounds good, marking this as resolved.

_docs-2/security/on-disk-encryption.md
cshannon: I'm adding some comments about this in my latest update but is there any advice we should give to people in this case?

cshannon:
I've made all the requested changes so this can be reviewed again and I can make any further updates.

dlmarion (approved):

PR #343, opened 2022-10-01
Update troubleshooting tools documentation

### LINKED ISSUE #337: Update troubleshooting tools documentation with information on new tools added in #2807
After apache/accumulo#2807 is merged the [documentation](https://accumulo.apache.org/docs/2.x/troubleshooting/tools) will need to be updated with instructions for the new tools that are being added to the `accumulo` command.

cshannon:
This commit adds updates for the new utilities that were added to the acuumulo accumulo admin command. Also add information on rfile-info when reading encrypted files.

This closes #337

### DIFF
--- a/_docs-2/troubleshooting/tools.md
+++ b/_docs-2/troubleshooting/tools.md
@@ -65,7 +65,7 @@ To examine an encrypted rfile the necessary encryption properties must be provid
 the `accumulo.properties` file can be copied, the necessary encryption parameters added, and then the properties file can 
 be passed to the utility with the `-p` argument. 
 
-For example, if using `PerTableCryptoFactory` and the `AESCryptoService`, you would need the following properties in
+For example, if using `PerTableCryptoServiceFactory` and the `AESCryptoService`, you would need the following properties in
 your accumulo.properties file:
 
 ```
--- a/_docs-2/troubleshooting/tools.md
+++ b/_docs-2/troubleshooting/tools.md
@@ -59,6 +59,41 @@ Likewise, `rfile-info` will dump the key-value pairs and show you the contents o
     row columnFamily:columnQualifier [visibility] timestamp deleteFlag -> Value
     ...
 
+### Encrypted Files
+
+To examine an encrypted rfile the necessary encryption properties must be provided to the utility. To do this
+the `accumulo.properties` file can be copied, the necessary encryption parameters added, and then the properties file can 
+be passed to the utility with the `-p` argument. 
+
+For example, if using `PerTableCryptoFactory` and the `AESCryptoService`, you would need the following properties in
+your accumulo.properties file:
+
+    general.custom.crypto.key.uri=<path-to-key>/data-encryption.key
+    instance.crypto.opts.factory=org.apache.accumulo.core.spi.crypto.PerTableCryptoServiceFactory
+    table.crypto.opts.service=org.apache.accumulo.core.spi.crypto.AESCryptoService
+
+Example output:
+
+    $ accumulo rfile-info /accumulo/tables/1/default_tablet/A0000004.rf -p <path-to-properties>/accumulo.properties
+    2022-10-01T12:19:32,564 [rfile.PrintInfo] WARN : Attempting to find file across filesystems. Consider providing URI instead of path
+    Reading file: hdfs://localhost:8020/accumulo/tables/1/default_tablet/A0000004.rf
+    Encrypted with Params: ...
+    2022-10-01T12:19:32,760 [bcfile.CompressionAlgorithm] INFO : Trying to load codec class org.apache.hadoop.io.compress.LzoCodec
+    2022-10-01T12:19:32,762 [bcfile.CompressionAlgorithm] INFO : Trying to load codec class org.apache.hadoop.io.compress.Lz4Codec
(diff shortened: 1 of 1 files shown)


cshannon:
This is ready for review but I made it a draft pull request for now because some of the output depends on https://github.com/apache/accumulo/pull/2992 being merged first.

_docs-2/troubleshooting/tools.md
```diff
@@ -59,6 +59,41 @@ Likewise, `rfile-info` will dump the key-value pairs and show you the contents o
...
+your accumulo.properties file:
+
+    general.custom.crypto.key.uri=<path-to-key>/data-encryption.key
+    instance.crypto.opts.factory=org.apache.accumulo.core.spi.crypto.PerTableCryptoServiceFactory
+    table.crypto.opts.service=org.apache.accumulo.core.spi.crypto.AESCryptoService
+
+Example output:
+
+    $ accumulo rfile-info /accumulo/tables/1/default_tablet/A0000004.rf -p <path-to-properties>/accumulo.properties
+    2022-10-01T12:19:32,564 [rfile.PrintInfo] WARN : Attempting to find file across filesystems. Consider providing URI instead of path
```
ctubbsii, line 78: I think it's better to use the triple backtick method to create blocks, rather than rely on indentation, which is less obvious, and has more limitations (like no syntax highlighting).

_docs-2/troubleshooting/tools.md
```diff
@@ -59,6 +59,41 @@ Likewise, `rfile-info` will dump the key-value pairs and show you the contents o
...
+    Reading file: hdfs://localhost:8020/accumulo/tables/1/default_tablet/A0000004.rf
+    Encrypted with Params: ...
+    2022-10-01T12:19:32,760 [bcfile.CompressionAlgorithm] INFO : Trying to load codec class org.apache.hadoop.io.compress.LzoCodec
+    2022-10-01T12:19:32,762 [bcfile.CompressionAlgorithm] INFO : Trying to load codec class org.apache.hadoop.io.compress.Lz4Codec
+    2022-10-01T12:19:32,765 [bcfile.CompressionAlgorithm] INFO : Trying to load codec class org.apache.hadoop.io.compress.ZStandardCodec
+    2022-10-01T12:19:32,767 [bcfile.CompressionAlgorithm] INFO : Trying to load codec class org.apache.hadoop.io.compress.DefaultCodec
```
ctubbsii, line 84: I'm not sure it's useful to show the user all the logging messages in the output, especially these about compression algorithms. They create noise, and distract from the tool behavior we are trying to document. Also, they depend heavily on the user's class path and configuration, so they may not experience the same output anyway. It'd probably be enough to just have a footnote that there may be log messages from the code that show up in the output, depending on their logging configuration.

_docs-2/troubleshooting/tools.md
```diff
@@ -59,6 +59,41 @@ Likewise, `rfile-info` will dump the key-value pairs and show you the contents o
...
+For example, if using `PerTableCryptoFactory` and the `AESCryptoService`, you would need the following properties in
+your accumulo.properties file:
+
+    general.custom.crypto.key.uri=<path-to-key>/data-encryption.key
+    instance.crypto.opts.factory=org.apache.accumulo.core.spi.crypto.PerTableCryptoServiceFactory
+    table.crypto.opts.service=org.apache.accumulo.core.spi.crypto.AESCryptoService
+
+Example output:
+
+    $ accumulo rfile-info /accumulo/tables/1/default_tablet/A0000004.rf -p <path-to-properties>/accumulo.properties
```
ctubbsii, line 77: The example should use the fully qualified URI, rather than just the path, in order to avoid the subsequent warning, and set a better example to users. Something like `hdfs://localhost:9020/accumulo/tables...`

_docs-2/troubleshooting/tools.md
cshannon: Good point, I will get rid of that as it is too verbose and there's also already a full example of output earlier.

cshannon:
Thanks for the quick feedback @ctubbsii, I should have some time Monday or Tuesday to make updates.

cshannon:
I made updates based on the feedback given. I will take another look Friday in more detail when I have more time and do anymore tweaks/updates and then this should be ready for a final review instead of draft status.

cshannon:
@ctubbsii - I don't think I have any more changes for now for this so it can be reviewed again as I think I addressed all the comments. Let me know if you see anything else.

milleruntime (approved):

_docs-2/troubleshooting/tools.md
```diff
@@ -59,6 +59,45 @@ Likewise, `rfile-info` will dump the key-value pairs and show you the contents o
     row columnFamily:columnQualifier [visibility] timestamp deleteFlag -> Value
     ...
 
+### Encrypted Files
+
+To examine an encrypted rfile the necessary encryption properties must be provided to the utility. To do this
+the `accumulo.properties` file can be copied, the necessary encryption parameters added, and then the properties file can 
+be passed to the utility with the `-p` argument. 
+
+For example, if using `PerTableCryptoFactory` and the `AESCryptoService`, you would need the following properties in
```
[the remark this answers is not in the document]
cshannon: Yep you are right that should be `PerTableCryptoServiceFactory`

ctubbsii (approved):

PR #344, opened 2022-10-13
Add Known Issue about VFS ClassLoader in latest release notes for each version

dlmarion:


ctubbsii (commented):
@dlmarion It looks like you and I were both working on a PR for the same thing at the same time. I created #345. I think what you have here is fine, but if there's anything you want to grab from mine, please feel free to do so, and merge it in with this.

PR #346, opened 2022-10-15
Add recent notable changes to 2.1.0 release notes

cshannon:


Manno15 (approved):

ctubbsii:
As an adjective, it should be "On-Disk Encryption". As a prepositional phrase clarifying the location of the encryption, it can be "Encryption On Disk". "On Disk Encryption" reads like you're talking *about* "Disk Encryption", generally. That may also be suitable in some cases, but I don't think it reads the best in this context.

ctubbsii (commented):
These are fine for now, but we'll probably trim quite a bit of this down before it's final, because we really want to call out the most important changes that will impact users, rather than enumerate our development progress.

However, this should be based on the main branch, not the next-release branch. We can prepare the draft release notes in place.

PR #347, opened 2022-10-15
Increase build timeout to 15 minutes

cshannon:
I've seen a couple of the QA builds timeout with only 10 minutes due to the test site part of the build taking a bit longer than usual. The jobs would have completed but timed out at 10 minutes (were around 80% done the upload part of the build).  Examples:

https://github.com/apache/accumulo-website/actions/runs/3256658816/attempts/1
https://github.com/apache/accumulo-website/actions/runs/3245360886

### DIFF
--- a/.github/workflows/jekyll.yaml
+++ b/.github/workflows/jekyll.yaml
@@ -27,7 +27,7 @@ on:
 
 jobs:
   jekyll:
-    timeout-minutes: 10
+    timeout-minutes: 15
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
@@ -54,4 +54,3 @@ jobs:
         name: site
         path: ./_site/
         if-no-files-found: ignore
-


ctubbsii:
2 weeks ago, the build only took 3-4 minutes total. Did something happen to change that?

ctubbsii:
Never mind, looking further back, it seems the build time has always fluctuated quite a bit, and has gone up to 10 minutes before.

.github/workflows/jekyll.yaml
```diff
@@ -27,7 +27,7 @@ on:
 
 jobs:
   jekyll:
-    timeout-minutes: 10
+    timeout-minutes: 15
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
```
ctubbsii, line 33: I wonder if some of this would be faster if we updated to use v3 for the actions. Could make that change as well (or use whatever is the latest, if it's newer than v3).

ctubbsii (changes requested):

cshannon:
> Never mind, looking further back, it seems the build time has always fluctuated quite a bit, and has gone up to 10 minutes before.

Yeah, I was going to respond and say I wasn't sure but I don't think anything changed. When I looked at why it failed and the build histories it there seems to be quite a bit of fluctuation. My guess is it's just related to being shared Apache servers so depending on server load and what is going on you get different build times.

PR #351, opened 2022-10-19
Add SVN instructions for updating the KEYS file

### LINKED ISSUE #350: Add svn instructions to Making a Release web page
The instructions say to upload the artifacts to a SVN repo [here](https://accumulo.apache.org/contributor/making-release#copy-artifacts-to-distapacheorg). Having not used svn in over a decade, it would be very useful to have example CLI instructions.

dlmarion:
Closes #350

### DIFF
--- a/contributor/making-release.md
+++ b/contributor/making-release.md
@@ -20,7 +20,19 @@ Follow these steps to make a release of Apache Accumulo.
 There are number of things that are required before attempting to build a release.
 
 1. Use gpg-agent, and be sure to increase the gpg-agent cache timeout (via .gnupg/gpg-agent.conf) to ensure that the agent doesn't require re-authentication mid-build, as it will cause things to fail. For example, you can add `default-cache-ttl 6000` to increase the timeout from the default of 10 minutes to over an hour. If you do not have a GPG key, reference the very thorough [ASF release signing documentation][1].
-2. Once you are able to sign a release, add your gpg key to the [KEYS file][keys-file] (instructions are at the top of the file).  This will require updating the SVN repository (more below).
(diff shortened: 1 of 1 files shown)


contributor/making-release.md
```diff
@@ -20,7 +20,19 @@ Follow these steps to make a release of Apache Accumulo.
...
+   sudo yum install -y git-svn
+   # clone the SVN repo into the directory accumulo-dist-gitsvn
+   git svn clone https://dist.apache.org/repos/dist/release/accumulo accumulo-dist-gitsvn
+   cd accumulo-dist-gitsvn
+   # make changes to KEYS file, then commit the changes locally
+   git add KEYS
+   git commit
+   # push the changes to the remote SVN repo
+   git svn dcommit --username=<ASF_Username>
+   ```
```
ctubbsii, line 35: I was worried that we were going to have a bunch of verbose documentation for svn, which I thought would be overwhelming and noisy. In general, I really hate adding documentation for other software/tools into our documentation, because it can quickly bloat our docs with scope creep. However, I think you did a good job of keeping it simple and to the point here.

Alternatively, using `svn` directly to edit only the KEYS file:

```bash
svn co https://dist.apache.org/repos/dist/release/accumulo --depth empty <checkoutDir>
svn up <checkoutDir>/KEYS
# edit the KEYS file
svn ci <checkoutDir>/KEYS
```

dlmarion: @ctubbsii  - did you want me to add your alternative?

ctubbsii: I think mine is a simpler example, as a bare minimum. But, I don't want our doc to be bloated with training people to be competent with tools that we aren't responsible for (those tools' own docs are best for that). So, if you think yours are better for most of us, go with that one. I'd rather not have two, because then where does it end? Should we provide instructions with TortoiseSVN, eSvn, etc.? Best to stick with one. I think the native svn CLI is the best example of bare minimum functionality, but more of our devs are used to git. Then again, git-svn is a whole different beast on top of git. So, I'll leave it to your discretion.

PR #352, opened 2022-10-20
Update 2.1.0 relnotes permission checks for configuration information

dlmarion:


EdColeman (approved):
Are the thrift changes more of an internal change?  Okay with the shell config command changes because that is user facing.

dlmarion:
> Are the thrift changes more of an internal change? Okay with the shell config command changes because that is user facing.

No, it affects the Thrift API, which in theory users could be calling directly.

ctubbsii:
> > Are the thrift changes more of an internal change? Okay with the shell config command changes because that is user facing.
> 
> No, it affects the Thrift API, which in theory users could be calling directly.

Yeah, they could... but that's not public API. We don't need to publish internal implementation details to users in these release notes. If somebody is using internal APIs, they should follow the internal development activity to stay abreast of those kinds of changes.

EdColeman (approved):

PR #354, opened 2022-10-24
Describe how to create the new host configuration file during upgrade

dlmarion:


_docs-2/administration/upgrading.md
```diff
@@ -30,6 +27,12 @@ after installing 2.1 but before starting it.
 ${ACCUMULO_HOME}/bin/accumulo org.apache.accumulo.manager.upgrade.RenameMasterDirInZK
 ```
 
+### Create new cluster configuration file
+
+The `accumulo-cluster` script now uses a single file that defines the location of the managers,
+tservers, etc. You can create this file using the command `accumulo-cluster create-config`. You
+will then need to transfer the contents of the individual files to this new consolidated file.
```
EdColeman, line 34: suggest```
will then need to transfer the contents of the current individual files to this new consolidated file.
```

EdColeman (approved):

PR #367, opened 2022-11-03
update upgrade instructions to include prop conversion

EdColeman:


dlmarion (approved):
This looks good, I think it will be really useful for people.

_docs-2/administration/upgrading.md
```diff
@@ -11,6 +11,24 @@ to 2.1 and then start 2.1. To upgrade from a release prior to 1.10, follow the
...
+includes transactions that have completed with `SUCCESS` or `FAILED` but have not been removed by the automatic clean-up 
+process.  This is required because the internal serialization of FATE transactions is not guaranteed to be compatible 
+between versions, so *ANY* FATE transaction that is present will fail the upgrade.  Procedures to manage FATE 
+transactions, including commands to fail and delete transactions, are included in the Accumulo documentation 
+[FATE - Administration]({% durl administration/fate#administration %})
+
+Two significant changes from 2.0 to 2.1 that are important to note: 1) properties and services that referenced
+`master` are renamed `manager` and 2) the property storage in ZooKeeper has changed from a ZooKeeper node per
+property to a single node for all properties with one ZooKeeper property node for each table, namespace and 
+the system config properties.
```
ctubbsii, line 24: This list of two items can be turned into a markdown list:

```suggestion
Two significant changes from 2.0 to 2.1 that are important to note:

1. properties and services that referenced `master` are renamed `manager` and
2. the property storage in ZooKeeper has changed from a ZooKeeper node per property  to a single node for all properties with one ZooKeeper property node for each table, namespace. and the system config properties.
```

_docs-2/administration/upgrading.md
```diff
@@ -38,6 +56,43 @@ will then need to transfer the contents of the current individual files to this
 **Warning**: Upgrading a previously encrypted instance is not supported as the implementation
 and properties have changed.
 
+## Upgrading from 2.0 to 2.1
+
+Upgrading to Accumulo 2.1 is done by stopping Accumulo 2.0 and starting Accumulo 2.1.
+Before stopping Accumulo 2.0, install Accumulo 2.1 and configure it by following the 
+[Quick Start]({% durl getting-started/quickstart %}).
```
ctubbsii, line 63: I'm not sure installing 2.1 always makes sense to do before stopping 2.0. In some cases, the version may be managed by an RPM, which will replace 2.0 with 2.1. Also, this is misleading, because it suggests doing the quickstart to set up a new 2.1 cluster... but that's not what you would want to do if you're upgrading.

_docs-2/administration/upgrading.md
```diff
@@ -38,6 +56,43 @@ will then need to transfer the contents of the current individual files to this
...
+
+Stop Accumulo.
+(optional) create a ZooKeeper snapshot using:
```
ctubbsii, line 66: Markdown will merge these lines on display. If you want a separate line, put a blank line in between. Or use a markdown list by prefixing each line with `* `.

_docs-2/administration/upgrading.md
```diff
@@ -38,6 +56,43 @@ will then need to transfer the contents of the current individual files to this
...
+
+`accumulo dump-zoo --xml --root /accumulo | tee PATH_TO_SNAPSHOT`
```
ctubbsii, line 68: This inline code will look funny on its own. If you want a code block, use triple quotes:

````suggestion
```
accumulo dump-zoo --xml --root /accumulo | tee PATH_TO_SNAPSHOT
```
````

File: _docs-2/administration/upgrading.md [hunk 5]
```diff
@@ -38,6 +56,43 @@ will then need to transfer the contents of the current individual files to this
...
+2022-11-03T14:35:45,813 [util.ConfigTransformer] INFO : property transform for TablePropKey{TableId=!0'} took 72ms ms, delete count: 26, error count: 0
+...
+```
+If the upgrade utility is not used, similar messages will be printed to the manager log on the first 2.1 version 
+manager start up.
+
+When the property conversion is complete, you can verify the configuration using the zoo-info-viewer utility 
+(new in 2.1)
+
+`accumulo zoo-info-viewer  --print-props`
```
ctubbsii, line 94:
````suggestion
```
accumulo zoo-info-viewer  --print-props
```
````

**Review by ctubbsii — CHANGES_REQUESTED** [review]

File: _docs-2/administration/upgrading.md (hunk 2 above)
Comment (EdColeman) [reply _docs-2/administration/upgrading.md:63]: The quick start reference was copied from the 1.8/9 to 2.0 section to be consistent.  I'll work on rewording, but there is a chicken / egg issue 

If installed from an rpm - would it be better to stop accumulo and then install the rpm or does it not matter?  I think it does, because if you want to use accumulo scripts to stop the running version, those scripts need to match the process names of that running version - but to get the config upgrade utility, you need 2.1 code.

Comment (ctubbsii) [reply _docs-2/administration/upgrading.md]: You could just as easily stop the 2.0 instance, install 2.1, run the upgrade utility, then start 2.1 instead of installing 2.1 prior to stopping 2.0.

The upgrade utility shouldn't run on a live 2.0 instance, but on a stopped one. So, I think what's important is that the optional config upgrade utility needs to be run after stopping the 2.0 instance, but before starting the 2.1 instance. I don't think it has anything to do with the order of installation relative to stopping the instance.

Comment (EdColeman) [reply _docs-2/administration/upgrading.md]: I am trying to allow for installations that have a sensitivity to the length of downtime time required for the upgrade process to complete.  The time required to shutdown, run the upgrade utilities and then restart is the minimum that is necessary and the upgrade utilities runtime should be small (maybe a minute?)  and overall the time is relatively quick.

It is common for people to request the ability to do a rolling upgrade just to avoid the down-time - but with these changes that is not possible, the configuration must be converted before tablets are assigned to tservers and supporting both schemes in a running cluster is not just not supported, it flat out will not work. And that's without FATE and Thrift changes. Running 2.0 and 2.1 processes together is just not possible.

The time to distribute the new files and install the configuration could be as long or longer as the complete restart part of the process.  Because the distribution and configuration can be done with the cluster running, it does not need to contribute to the down time.  The key part being that the upgrade utilities need to be run with 2.1 code on a stopped 2.0 instance.

It all depends on the comfort level and the requirements of the user.  I do not want to imply that the cluster must be shutdown before installing the 2.1 software. But I also did not want to imply that someone needs to have a parallel installation of the software with 2.0 and 2.1 binaries simultaneously either.

My metal model is along the lines of preferring a  java-alternatives model where multiple versions are installed and available and selected with environment variables.  Probably a bad analogy because once running, we do not support switching back, but it also seems to be the best way to support restoring from a ZooKeeper dump and continuing with 2.0 if things went unexpectedly off-the rails and they did not want to extend the downtime while things get sorted.

File: _docs-2/administration/upgrading.md [hunk 6]
```diff
@@ -38,6 +56,60 @@ will then need to transfer the contents of the current individual files to this
 **Warning**: Upgrading a previously encrypted instance is not supported as the implementation
 and properties have changed.
 
+## Upgrading from 2.0 to 2.1
+
+Note, the process fro upgrading to 2.1 assumes that you have already upgraded to 2.0.  Upgrading from a version prior
```
dlmarion, line 61:
```suggestion
Note, the process for upgrading to 2.1 assumes that you have already upgraded to 2.0.  Upgrading from a version prior 
```

_docs-2/administration/upgrading.md
```diff
@@ -38,6 +56,60 @@ will then need to transfer the contents of the current individual files to this
 **Warning**: Upgrading a previously encrypted instance is not supported as the implementation
 and properties have changed.
 
+## Upgrading from 2.0 to 2.1
```
dlmarion, line 59: Should the title here be "Upgrading from 1.10 or 2.0 to 2.1"? Upgrading from 1.10 is supported and tested IIRC

_docs-2/administration/upgrading.md
```diff
@@ -38,6 +56,60 @@ will then need to transfer the contents of the current individual files to this
...
+
+Note, the process fro upgrading to 2.1 assumes that you have already upgraded to 2.0.  Upgrading from a version prior 
+to 2.0 may work, but is unsupported and untested.  The basic idea is to:
```
dlmarion, line 62: This statement doesn't agree with the upgrading from 1.10.

_docs-2/administration/upgrading.md
```diff
@@ -38,6 +56,60 @@ will then need to transfer the contents of the current individual files to this
...
+- (optional) create a ZooKeeper snapshot
+- (optional) run the configuration upgrade utility
+- start Accumulo 2.1
+
+The particular order for your installation may differ depending on how you choose to install Accumulo. You may want
+to stop Accumlo first if your installation is going to over-write your current installation.  The important part is that
+Accumulo 2.1 manager does not start running before the 2.1 configuration is in place for all of your nodes and if you
+want to run the optional steps to create a ZooKeeper snapshot and run the configuration update stand-alone utility.
+
+### Create ZooKeeper snapshot (optional)
```
dlmarion, line 76:
```suggestion
### Create ZooKeeper snapshot (optional, but recommended)
```

dlmarion (changes requested):

_docs-2/administration/upgrading.md
EdColeman: Is it our official policy to support upgrades from 1.10 to 2.1?  I was not sure and modeled the text to our past statements that we do not support multiple leaps with one upgrade.  I'm fine if that's true and my internal testing usually goes from 1.10 to 2.1 - but that's using uno so I was uncertain if there configuration / property file settings in a cluster that could cause issues.  Functionally, 1.10 to 2.1 is fine - its the supporting scripts / config files that I am fuzzy on.

dlmarion: IIRC, in discussions with Christopher, upgrades from 1.10 to 2.1 should work.

EdColeman: I do not dispute they work.  But do we want to officially declare that?  I'll change the text to reflect that - we can modify it if that is not our official position.

_docs-2/administration/upgrading.md
ctubbsii: If you're trying to minimize the downtime, you could reword to achieve that goal, but in a way that's less misleading about the order requirement. Something along the lines of: "To minimize downtime, if your installation method allows for it, prepare the 2.1 binaries before you shut down 2.0, so you can upgrade and start 2.1 right away."

ctubbsii: > The key part being that the upgrade utilities need to be run with 2.1 code on a stopped 2.0 instance.

I think it would be good to emphasize this, perhaps word-for-word.

_docs-2/administration/upgrading.md
ctubbsii: Yes, we do want to officially declare that. The two supported upgrade paths are: `LTM.latest --> (LTM+1).latest` and `major.minor.latest --> major.(minor+1).latest`. So, for here, we support 2.0.1 to 2.1.0, and 1.10.2 to 2.1.0. However, the upgrade instructions might be slightly different for 1.10 than for 2.0.

_docs-2/administration/upgrading.md
```diff
@@ -38,7 +57,68 @@ will then need to transfer the contents of the current individual files to this
...
+The particular order for your installation will differ depending on how you choose to install Accumulo.  
+The key part being that the upgrade utilities need to be run with 2.1 code on a stopped 1.10 / 2.0 instance. 
+
+You can elect to stop your current instance, install and configure the 2.1 binaries, optionally run the upgrade 
+utilities and then start 2.1, or you could install and configure the 2.1 binaries while your instance remains 
+running and then when ready, stop your instance, optionally run the upgrade utilities and then start 2.1.
+
+The basic upgrade sequence is to: 
+
+- stop Accumulo 2.0
```
dlmarion, line 78:
```suggestion
- stop Accumulo
```

dlmarion (approved):
Looks good, one small nit.

Manno15 (approved):

_docs-2/administration/upgrading.md
```diff
@@ -38,7 +57,70 @@ will then need to transfer the contents of the current individual files to this
 **Warning**: Upgrading a previously encrypted instance is not supported as the implementation
 and properties have changed.
 
-## Upgrading from 1.8/9 to 2.0
+## Upgrading from 1.9/1.10/2.0 to 2.1
```
ctubbsii, line 60: This section header is redundant. The one above says "Upgrading from 1.10.x or 2.0.x to 2.1"
These sections need to be consolidated. As-is, it's very noisy and confusing, as the information seems spread out between the sections.

The only upgrade paths to 2.1 that we support are: 1.10.latest to 2.1.latest and 2.0.latest to 2.1.latest
Any information about versions earlier than 1.10 should simply say to upgrade to 1.10.latest first. Versions earlier than 1.10 have been EOL for nearly 2 years, I think. I don't think we need to waste time on even mentioning them here. We can direct people to the user list for help if they need to update to 1.10.

_docs-2/administration/upgrading.md
```diff
@@ -38,7 +57,70 @@ will then need to transfer the contents of the current individual files to this
...
+
+Note: the process for upgrading to 2.1 assumes that you have already upgraded to 1.10 or 2.0.  Upgrading from a 
+version prior to 1.10 may work, but is unsupported and untested.  If you are upgrading from 1.10 make sure
```
ctubbsii, line 63: Lots of things "may" work. I don't think we need to speculate on those scenarios. Also, we know that *most* versions prior to 1.10 definitely won't work, so this language is really sub-optimal. We should focus on the supported upgrade paths, and direct people to the mailing list for help with anomalous scenarios.

ctubbsii (changes requested):

_docs-2/administration/upgrading.md
EdColeman: This was trying to hedge that it seems likely that because the changes between 1.9.3 and 1.10 are small it seems likely that upgrading from 1.9.3 to 2.10 would probability work - but I have not tried it.

_docs-2/administration/upgrading.md
ctubbsii: I'll fix this in a subsequent commit

PR #380, opened 2023-03-31
Fixes #379 creates emeritus PMC table on people page

### LINKED ISSUE #379: Provide space on "people" page for Emeritus PMC members
At least one person on our PMC is now Emeritus (for that role, but remains a committer). The "people" page needs to be updated to be able to mark such individuals as Emeritus. Right now, it lists all PMC and committers in one big list. There are probably several different design options. Maybe using datatables could work to organize the list of contributors/committers/PMC members/Emeritus PMC members? Or just a separate "Committers / Emeritus PMC Members" section.

If anybody is interested in doing this, they could also suggest a similarly styled update over on the Fluo project for https://github.com/apache/fluo-website/issues/216

keith-turner:


pages/people.md
```diff
@@ -62,6 +60,14 @@ $(function() {
 | vikrams       | Vikram Srivastava                                 | [Cloudera][CLOUDERA]                   | [PT][PT] |
 | vines         | John Vines                                        | [sqrrl][SQRRL]                         | [ET][ET] |
 
+## Emeritus PMC and Committers
```
ctubbsii, line 63: I'm worried that this implies "Emeritus PMC" and "Emeritus Committers", when it's really "Emeritus PMC, but still current Committers". I can't think of a good way to communicate that succinctly in a section header. I thought about reversing the order, as in "Committers and Emeritus PMC" or "Committers Only (PMC Emeritus)", but then what if we have somebody who accepts an invite as committer, but declines being PMC. Then, they are "Committer only, never PMC". Will we want to have a separate section for that case? Or do we want to include them in this same section with an appropriate section header for both situations (and then mark inside the table that they are Emeritus for PMC).

I'm not sure which direction is best... but I'd prefer not to imply that they are Emeritus Committers when their Committer privileges are still held, and only the PMC status is Emeritus.

keith-turner: > I'm worried that this implies "Emeritus PMC" and "Emeritus Committers"

I agree, tt can definitely read that way.

> I can't think of a good way to communicate that succinctly

Looking at what you suggested, I like "Committers Only (PMC Emeritus)" and its only a few bytes longer so it still succinct. 

> but then what if we have somebody who accepts an invite as committer, but declines being PMC

We can worry about that case in the future when it happens.  We could create another section or maybe add a column to the table.

ctubbsii (approved):

pages/people.md
keith-turner: I pushed an update to the section name in 17591e0595bf823896807dd6c9fd41291a035833

ctubbsii:
After this was merged, I noticed a minor issue with reusing the table element ID for both tables. I fixed this in #381

PR #382, opened 2023-04-05
add confluence link to how-to-contribute page

EdColeman:
Update the how to contribute page to include a link to the Accumulo Confluence Space

### DIFF
--- a/pages/contact-us.md
+++ b/pages/contact-us.md
@@ -54,6 +54,11 @@ follow discussions as they happen.
 
 Drop by and chat about Accumulo at the [#accumulo][accumulo-channel] channel on the [ASF Slack workspace][asf-slack]. Join using this [invite link][slack-invite]. If you have trouble joining, email `dev@accumulo.apache.org` for an invite.
 
+## Confluence
+
+The [Accumulo Confluence Space][confluence] provides a collaboration space for design documents.  If you would like write 
+access, please send a request to `dev@accumulo.apache.org` with your Confluence username.
+
 ## Contributions
 
 Contributions to Apache Accumulo are welcome! If you are interested, read [how to contribute][contribute]. If you need help finding something to work on, send a message to our `dev` mailing list and we'll help you find a task that interests you.
@@ -94,3 +99,4 @@ Contributions to Apache Accumulo are welcome! If you are interested, read [how t
 [accumulo-channel]: https://the-asf.slack.com/messages/CERNB8NDC
 [slack-invite]: https://s.apache.org/slack-invite
 [contribute]: /how-to-contribute/
+[confluence]: https://cwiki.apache.org/confluence/display/ACCUMULO/Apache+Accumulo+Home
--- a/pages/contact-us.md
+++ b/pages/contact-us.md
@@ -54,11 +54,6 @@ follow discussions as they happen.
 
 Drop by and chat about Accumulo at the [#accumulo][accumulo-channel] channel on the [ASF Slack workspace][asf-slack]. Join using this [invite link][slack-invite]. If you have trouble joining, email `dev@accumulo.apache.org` for an invite.
 
-## Confluence
-
-The [Accumulo Confluence Space][confluence] provides a collaboration space for design documents.  If you would like write 
-access, please send a request to `dev@accumulo.apache.org` with your Confluence username.
-
 ## Contributions
 
 Contributions to Apache Accumulo are welcome! If you are interested, read [how to contribute][contribute]. If you need help finding something to work on, send a message to our `dev` mailing list and we'll help you find a task that interests you.
--- a/pages/contact-us.md
+++ b/pages/contact-us.md
@@ -94,4 +94,3 @@ Contributions to Apache Accumulo are welcome! If you are interested, read [how t
 [accumulo-channel]: https://the-asf.slack.com/messages/CERNB8NDC
 [slack-invite]: https://s.apache.org/slack-invite
 [contribute]: /how-to-contribute/
-[confluence]: https://cwiki.apache.org/confluence/display/ACCUMULO/Apache+Accumulo+Home
--- a/pages/how-to-contribute.md
+++ b/pages/how-to-contribute.md
@@ -31,11 +31,6 @@ Any questions/ideas don't hesitate to [contact us][contact].
 | [Proxy][p]         | [Issues][pi]                   | Apache Thrift service that exposes Accumulo to other languages |
 | [Maven plugin][m]  | [Issues][mi]                   | Maven plugin that runs Accumulo |
 
-## Confluence
-
-The [Accumulo Confluence Space][confluence] provides a collaboration space for design documents.  If you would like write
-access, please send a request to `dev@accumulo.apache.org` with your Confluence username.
-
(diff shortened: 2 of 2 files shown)


keith-turner (approved):

pages/contact-us.md
```diff
@@ -54,6 +54,11 @@ follow discussions as they happen.
 
 Drop by and chat about Accumulo at the [#accumulo][accumulo-channel] channel on the [ASF Slack workspace][asf-slack]. Join using this [invite link][slack-invite]. If you have trouble joining, email `dev@accumulo.apache.org` for an invite.
 
+## Confluence
+
+The [Accumulo Confluence Space][confluence] provides a collaboration space for design documents.  If you would like write 
+access, please send a request to `dev@accumulo.apache.org` with your Confluence username.
+
```
ctubbsii, line 61: I don't think the contact page is the appropriate place for this. I think it would be better to mention the use of Confluence on a page like https://accumulo.apache.org/how-to-contribute/, maybe under the "Helpful Links" section, or maybe in its own section referring to developer design documentation. It should link back to the contact-us page instead of directly linking the mailing list name (just to keep it listed in one place, in case it changes).

The how to contribute page is actually a bit sparse. Awhile back, somebody on the ComDev mailing list suggested standardizing on projects creating a `/contribute` entry point to collect project-specific information on contributing. I think our "how to contribute" page kind of does that, but could perhaps use a bit of a refresher, including adding this info about Confluence.

EdColeman: Added a reference on how-to-contribute and a link back to the contact-us.

Being that the Confluence space is a communications channel and one way to interact with the Accumulo community, the contact-us page seems a logical place to look if someone is searching for a way to interact with the community.

The intention of this PR was to raise the visibility of the Confluence space. The refresh of the `/contribute` page can be done as a separate effort.

ctubbsii (changes requested):
I think the contact page should be reserved for information on how to contact us. Developer resources should be listed in the appropriate developer docs. We do not want people using Confluence as a means to contact us, and we don't want to raise visibility of Confluence as a means to contact us to non-developers (this was one of the main arguments in favor of having Confluence in the first place... to have an "internal" space for our developers to collaborate on design docs).

pages/contact-us.md
EdColeman: addressed in 110837d0

pages/how-to-contribute.md
```diff
@@ -8,6 +8,7 @@ Contributions are welcome to all Apache Accumulo repositories. While most contri
 there are other ways to contribute to Accumulo:
 
 * communicate on one of the [mailing lists](/contact-us/#mailing-lists)
+* participate in design discussions on [Confluence][confluence]
```
ctubbsii, line 11: I would not characterize Confluence as a place for "discussions". The mailing list is for that. Confluence is for drafting/posting/commenting-on design docs.

I suggest:

```suggestion
* view/draft/comment on designs or other docs in [Confluence][confluence]
```

pages/how-to-contribute.md
```diff
@@ -19,16 +20,16 @@ Any questions/ideas don't hesitate to [contact us][contact].
...
+| Repository         | Links                          | Description |
+| -------------------| ------------------------------ | ----------- |
+| [Accumulo][a]      | [Contribute][ac] [Issues][ai]  | Core Project |
+| [Website][w]       | [Contribute][wc] [Issues][wi]  | Source for this website |
+| [Examples][e]      | [Contribute][ec] [Issues][ei]  | Example code |
+| [Testing][t]       | [Contribute][tc] [Issues][ti]  | Test suites such as continuous ingest and random walk |
+| [Docker][d]        | [Contribute][dc] [Issues][di]  | Source for Accumulo Docker image |
+| [Wikisearch][s]    | [Contribute][sc] [Issues][si]  | Example application that indexes and queries Wikipedia data |
+| [Proxy][p]         | [Issues][pi]                   | Apache Thrift service that exposes Accumulo to other languages |
+| [Maven plugin][m]  | [Issues][mi]                   | Maven plugin that runs Accumulo |
```
ctubbsii, line 32: Since these are optional, I would take it or leave it. I don't think it matters, as long as it's consistent. I think I recently fixed a few lines on the people page that were missing these, but only because all the other lines had them, not because they are needed.

I'm okay if you leave this change in, but it's also fine to leave it unchanged. I don't think it matters.

pages/how-to-contribute.md
```diff
@@ -62,10 +63,13 @@ Any questions/ideas don't hesitate to [contact us][contact].
 
 * **Build resources** - [Jenkins][jenkins]
 * **Releases** - [Making a release][making], [Verifying a release][verifying]
+* **Confluence** - [Accumulo Confluence Space][confluence] provides a collaboration space for design documents.  If you would like write
+access, please send a request to `dev@accumulo.apache.org` with your Confluence username.
```
ctubbsii, line 67: I would characterize this as a "require" rather than a "want", because there should be an actual need. Characterizing it as a need, implies that the requester provide some answer to "why?" up front, rather than us asking it as a follow-up question. So, instead of "I want write access" (and us having to ask "for what?"), it becomes "I require write access because I want to contribute to document X".

```suggestion
* **Confluence** - [Accumulo Confluence Space][confluence] provides a collaboration space for design documents. If you require write
access, please [contact us][contact] with your Confluence username and request details.
```

ctubbsii (changes requested):

pages/how-to-contribute.md
EdColeman: This resolves a warning about possible incompatibility between versions - so it seemed safer to include them, even if optional.

ctubbsii: Sounds good. What warning, though? Was this in the Jekyll build output, or in a Markdown-aware text editor you use?

ctubbsii (approved):

pages/how-to-contribute.md
EdColeman: IDE.  

```
For compatibility reasons all table rows should have borders (pipe symbols) at the start and at the end
```

ctubbsii: For what it's worth, we specifically use Kramdown for GitHub-flavored-markdown (gfm), which is what Jekyll uses. So we don't really require compatibility with restrictive variants of Markdown that the IDE is trying to warn about. But, if it gets rid of the IDE error, I'm fine with it. I don't ever use an IDE for Markdown (or for the website at all). I'm a little surprised anybody does. I suspect it might be a bit challenging to get it working well in an IDE, given the mix of Ruby, Bundler, Jekyll, HTML, CSS, JavaScript, Markdown, Liquid, YAML, and who knows how many other languages are mixed in that I missed. :smiley_cat:

EdColeman: Well, for the spell check alone I find it worth it.  Does not help when I end up with a correctly spelled, but incorrect word but maybe chatGPT will come to my rescue someday.

IntelliJ does provide a preview pain that does give a good rough idea of what the render looks like - some things for our pages don't work, but what does helps getting a rough idea of how things are rendered, tables, font sizes and highlighting which I do find useful.

PR #388, opened 2023-07-27
Update webrick version

ddanielr:
Updating webrick version from 1.7.0  to latest version available.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,20 +1,21 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    addressable (2.8.1)
+    addressable (2.8.4)
       public_suffix (>= 2.0.2, < 6.0)
     colorator (1.1.0)
-    concurrent-ruby (1.1.10)
+    concurrent-ruby (1.2.2)
     em-websocket (0.5.3)
       eventmachine (>= 0.12.9)
       http_parser.rb (~> 0)
     eventmachine (1.2.7)
     ffi (1.15.5)
     forwardable-extended (2.6.0)
+    google-protobuf (3.23.4-x86_64-linux)
     http_parser.rb (0.8.0)
-    i18n (1.12.0)
+    i18n (1.14.1)
       concurrent-ruby (~> 1.0)
-    jekyll (4.3.1)
+    jekyll (4.3.2)
       addressable (~> 2.4)
       colorator (~> 1.0)
       em-websocket (~> 0.5)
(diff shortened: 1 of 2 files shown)


DomGarguilo:
The changes in `Gemfile.lock` were created automatically, right? Also were you able to verify that things still work with these changes?

ctubbsii:
I'm not sure what webrick is even being used for. It was added in 4980f0d0519cdd02d17e5541f57590f093c8ee37 as part of an update by @jmark99 to use a JShell version of the tour, but I'm not sure what it's actually needed for.

I did a quick reset of my local ruby and gem installation and did a bundle install to ensure everything was updated, then I verified the generated differences were not significant. I did that commit in 17248a992b04e516d004e57ca16cbaad6ab5e628

Gemfile.lock
```diff
@@ -55,7 +55,7 @@ GEM
     rexml (3.2.6)
     rouge (4.1.2)
     safe_yaml (1.0.5)
-    sass-embedded (1.64.1-x86_64-linux-gnu)
+    sass-embedded (1.63.6-x86_64-linux-gnu)
```
ctubbsii, line 58: This looks like a downgrade. Probably just because you created this branch before I did the more recent `bundle update`.

ddanielr:
> The changes in `Gemfile.lock` were created automatically, right? Also were you able to verify that things still work with these changes?

yes they were created automatically once I updated the webrick version.
I was able to successfully build the containerized dev environment with no issues. I was then able to render and host the website files.

@ctubbsii I think it might be a dependency for Jekyll? 
To test this, I removed webrick from the Gemfile, deleted my Gemfile.lock file and then did a bundle install. 
The contents of the new Gemfile.lock still list webrick 1.8.1.

ctubbsii (commented):
After doing a general `bundle update`, webrick was updated to 1.8.1, the same version this was mandating. After resolving conflicts with this PR, I can see that there are no other substantial changes in this PR, so I think we can close it.

I verified that there were no substantial changes in the generated HTML output with the `bundle update` commit I added in 17248a992b04e516d004e57ca16cbaad6ab5e628.

The main thing that remains in this PR is setting the minimum resolved version to 1.8.1 instead of 1.7, but by using the syntax `~> 1.8.1` instead of `~> 1.8`, we'd be forcing it to stay on versions `< 1.9` instead of `< 2.0`, as it is now. I can't predict the future, so I don't know what those versions might eventually contain, so I can't really say which is better. But also, it doesn't really matter, because they both resolve to 1.8.1 right now, and that's locked in with the Gemfile.lock file. So, I don't think there's anything we need to keep from this PR and can close it.

In future, I recommend updating gems with `bundle update` when updating, rather than target specific gems to update. Also, if you know what webrick is actually doing for us, it'd be great to add that in a comment in the `Gemfile`. The jekyll ones are obvious, but this one is not.

ctubbsii:
Resolved in 17248a992b04e516d004e57ca16cbaad6ab5e628

PR #389, opened 2023-07-27
Update Docker image

ddanielr:
Updates the source docker image to match the version of ruby defined in the Gemfile.lock file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # making updates to the accumulo website without requiring the dev
 # to maintain a local ruby development environment.
 
-FROM ruby:2.7.8-slim-bullseye as base
+FROM ruby:3.2.2-slim-bullseye as base
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
     build-essential \


ddanielr:
This change has been tested and just brings the container up to date with ruby 3.2.2
https://github.com/apache/accumulo-website/blob/17248a992b04e516d004e57ca16cbaad6ab5e628/Gemfile.lock#L74

ctubbsii (approved):
LGTM. But, please know that version is only in the Gemfile because it's the version I had on my Fedora 38 dev machine when I last did a bundle update. I'm not sure what version is in the ASF runners that built the asf-site branch, or what other devs are using. Otherwise, I'd suggest updating the Gemfile to require using a newer minimum version as well.

ctubbsii:
Looks like Ubuntu 18.04 is used by the ASF INFRA-provided builders, which has Ruby 2.5.1. So, we can't bump the minimum unless we changed things to create a GitHub Actions based builder that commits to the asf-site branch. Not sure if we want to do that much work. I'm not sure what INFRA future plans are.

PR #390, opened 2023-08-02
update fate command moved from shell to admin

EdColeman:
The fate command was removed from the shell and now uses the admin command. Update the command usage examples in the advanced troubleshooting section.

### DIFF
--- a/_docs-2/troubleshooting/advanced.md
+++ b/_docs-2/troubleshooting/advanced.md
@@ -343,7 +343,7 @@ For versions 2.1 and later, you can use the admin fate command to delete complet
 
 If any of the operations are not complete, you should rollback the upgrade and troubleshoot completing them with your prior version.
 
-Prior to 2.1, the same step apply, but the fate command is accessed with the Accumulo shell instead of the admin command.  The shell commands are as follows: 
+Prior to 2.1, the same steps apply, but the fate command is accessed with the Accumulo shell instead of the admin command.  The shell commands are as follows: 
 
 use `fate print` to list transactions
 use `fate delete` to delete completed transactions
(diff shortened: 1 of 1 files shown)


dlmarion (approved):

_docs-2/troubleshooting/advanced.md
```diff
@@ -333,12 +333,12 @@ overridden by using the `--local-wal-directories` option on the tool. It can be
...
 
-You can use the shell to delete completed FATE transactions using the following:
+You can use the admin fate command to delete completed FATE transactions using the following:
 
 * Start tservers
 * Start shell
-* Run `fate print` to list all
-* If completed, just delete with `fate delete`
+* Run `accumulo admin fate --print` to list all transactions
+* If the transactions have completed, just delete with them with `accumulo admin fate --delete TXID [TXID...]`
```
ctubbsii, line 341: Does this work in all 2.x versions? These docs cover all 2.x. If it only works for a subset of these versions, it should be called out, as in "Prior to 2.x.y, the command was in the shell as ..." or similar.

EdColeman: It was updated in 2.1 with PR #2914

EdColeman: Addressed in f6ec66fb

_docs-2/troubleshooting/advanced.md
```diff
@@ -333,16 +333,21 @@ overridden by using the `--local-wal-directories` option on the tool. It can be
...
 * Start managers once there are no more fate operations
 
 If any of the operations are not complete, you should rollback the upgrade and troubleshoot completing them with your prior version.
 
+Prior to 2.1, the same step apply, but the fate command is accessed with the Accumulo shell instead of the admin command.  The shell commands are as follows:
```
ctubbsii, line 346:
```suggestion
Prior to 2.1, the same steps apply, but the fate command is accessed with the Accumulo shell instead of the admin command.  The shell commands are as follows: 
```

ctubbsii (approved):
small typo, otherwise LGTM

PR #393, opened 2023-08-15
Improvements to 2.1.2 release notes

DomGarguilo:
This PR:
* changes the note about the manager page changes in the monitor
* makes sure all bullet points have a period at the end
* slightly adjusts some misc. wordage


ctubbsii:
The discussion Ed was referring to wasn't a public discussion... just me privately expressing my opinion that I think updates to draft release notes don't really need a PR. As long as it's properly marked as draft, and doesn't break the website, I think updating its content is something you can just lean on the "commit then review" mindset more than the "PR just in case" mindset. You can still create PRs if you want... but the draft release notes are going to receive a bunch of updates/polishing/review before the draft label gets removed anyway. So, unless you want somebody to specifically check your wording for technical correctness, it's probably not a problem to just commit it directly.

Issue #394, opened 2023-08-21
Update javascript and css resources


ctubbsii:
Several libraries we use for the website are a bit out of date and could be updated. Here's a list of some:

* [ ] Bootstrap (currently using 3.3.7)
* [x] Font Awesome (currently using 4.0.3)
* [x] DataTables (currently using 1.10.12 based on jQuery 2.2.3)
* [x] jQuery (currently using 2.2.4)

Edit:
The above dependencies are all imported in default.html. The search page imports its own dependencies that need updating:

- [x] lunr
- [x] jquery

ctubbsii:
Font Awesome, DataTables, and jQuery were updated in #396

DomGarguilo:
lunr was updated in #399 and jquery import was removed from that page so the main import of jquery would be used.

PR #396, opened 2023-08-22
Update fontawesome, jquery and datatables versions

DomGarguilo:
This PR:
* bumps fontawesome from 4.0.3 -> 6.4.2
  * this required some changes to the class names of the icons
  * [details on the fontawesome site](https://fontawesome.com/docs/web/setup/upgrade/upgrade-from-v4#what-s-changed-from-version-4-to-6) outlining what has changed between these versions
  * the only visual change that took place was the link icon from https://fontawesome.com/v4/icon/link to https://fontawesome.com/icons/link?f=classic&s=solid
  * In the user manuals from 1.x versions, the html generated by asciidoc has a few places that font awesome is used and it looks like the old version but i am not sure where that code is generated or if we want to change that.
* bumps datatables from 1.10.12 to 1.13.6
  * the 1.10.12 import included a version of jquery with the download but I saw a note [on the downloads page](https://datatables.net/download/index) in the select packages section, that says "DataTables requires jQuery. Don't select either version if you already have it.". Because of this, the new link to datatables download does not include a version of jquery with it
* bumps jquery from 2.2.4 to 3.7.0

This partially addresses #394. Everything but the bootstrap update is handled here in this PR.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -21,15 +21,15 @@
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <link href="https://maxcdn.bootstrapcdn.com/bootswatch/3.3.7/paper/bootstrap.min.css" rel="stylesheet" integrity="sha384-awusxf8AUojygHf2+joICySzB780jVvQaVCAt1clU3QsyAitLGul28Qxb2r1e5g+" crossorigin="anonymous">
-<link href="//netdna.bootstrapcdn.com/font-awesome/4.0.3/css/font-awesome.css" rel="stylesheet">
-<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/v/bs/jq-2.2.3/dt-1.10.12/datatables.min.css">
+<link href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.4.2/css/all.min.css" rel="stylesheet">
+<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/v/dt/dt-1.13.6/datatables.min.css">
 <link href="{{ site.baseurl }}/css/accumulo.css" rel="stylesheet" type="text/css">
 
 <title>{% if page.title_prefix %}{{ page.title_prefix | escape }}{% endif %}{% if page.title %}{{ page.title | escape }}{% else %}{{ site.title | escape }}{% endif %}</title>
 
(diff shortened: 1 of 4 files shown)


ctubbsii (approved):
Looks good! Thanks @DomGarguilo !

_layouts/default.html
```diff
@@ -21,15 +21,15 @@
...
+<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/v/dt/dt-1.13.6/datatables.min.css">
 <link href="{{ site.baseurl }}/css/accumulo.css" rel="stylesheet" type="text/css">
 
 <title>{% if page.title_prefix %}{{ page.title_prefix | escape }}{% endif %}{% if page.title %}{{ page.title | escape }}{% else %}{{ site.title | escape }}{% endif %}</title>
 
-<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/2.2.4/jquery.min.js" integrity="sha256-BbhdlvQf/xTY9gja0Dq3HiwQF8LaCRTXxZKRutelT44=" crossorigin="anonymous"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.7.0/jquery.min.js" integrity="sha256-2Pmvv0kuTBOenSvLm6bvfBSSHrUJ+3A7x6P5Ebd07/g=" crossorigin="anonymous"></script>
 <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js" integrity="sha384-Tc5IQib027qvyjSMfHjOMaLkfuWVxZxUPnCJA7l2mCWNIpG9mGCD8wGNIcPD7Txa" crossorigin="anonymous"></script>
-<script type="text/javascript" src="https://cdn.datatables.net/v/bs/jq-2.2.3/dt-1.10.12/datatables.min.js"></script>
+<script type="text/javascript" src="https://cdn.datatables.net/v/dt/dt-1.13.6/datatables.min.js"></script>
```
ctubbsii, line 32: When Bootstrap is updated, this could switch over to use the Bootstrap 5 styling instead of the default DataTables styling.

ctubbsii:
> * In the user manuals from 1.x versions, the html generated by asciidoc has a few places that font awesome is used and it looks like the old version but i am not sure where that code is generated or if we want to change that.

Those were generated by the Maven build in the older Accumulo releases, and should stay as they are. They don't need to be updated.

_layouts/default.html
ctubbsii: Created #400 to switch to bootstrap style. Will need to update these again when updating to bootstrap 5.

PR #398, opened 2023-08-29
Upgrade bootstrap version from 3 to 5

DomGarguilo:
Partially addresses #394 

Upgrades the bootstrap version from 3.3.7 to 5.3.1

A lot of classes have changed between versions and some new css was added to maintain the styling after upgrading to version 5.

Here are some screenshots showing various pages before and after these changes. I will replace each image with a new screenshot if I end up making more changes.

Current:
Screenshot from 2023-08-28 16-28-25
New with these changes:
Screenshot from 2023-08-28 16-23-53

Current:
Screenshot from 2023-08-28 16-28-36
New with these changes:
Screenshot from 2023-08-28 16-24-06

Current:
Screenshot from 2023-08-28 16-25-28
New with these changes:
Screenshot from 2023-08-28 16-21-33

Current:
Screenshot from 2023-08-28 16-25-56
New with these changes:
Screenshot from 2023-08-28 16-21-40

Current:
Screenshot from 2023-08-28 16-26-05
New with these changes:
Screenshot from 2023-08-28 16-21-43

Current:
Screenshot from 2023-08-28 16-26-25
New with these changes:
Screenshot from 2023-08-28 16-22-08

Current:
Screenshot from 2023-08-28 16-26-52
New with these changes:
Screenshot from 2023-08-29 12-55-39

Current:
Screenshot from 2023-08-28 16-27-02
New with these changes:
Screenshot from 2023-08-29 12-55-46

Current:
Screenshot from 2023-08-28 16-27-13
New with these changes:
Screenshot from 2023-08-29 13-25-39

Current:
Screenshot from 2023-08-28 16-27-39
New with these changes:
Screenshot from 2023-08-28 16-23-00

Current:
Screenshot from 2023-08-28 16-28-02
New with these changes:
image

Current:
Screenshot from 2023-08-29 13-35-36
New with these changes:
Screenshot from 2023-08-29 15-10-09


pages/contact-us.md
```diff
@@ -45,10 +45,10 @@ follow discussions as they happen.
...
-| **commits**       | Code changes                                     | [<span class="glyphicon glyphicon-search"/> Archive][C_A] | [<span class="glyphicon glyphicon-plus"/> Subscribe][C_SU] [<span class="glyphicon glyphicon-remove"/> Unsubscribe][C_UN] | |
-| **notifications** | Automated notifications (GitHub, etc.)             | [<span class="glyphicon glyphicon-search"/> Archive][N_A] | [<span class="glyphicon glyphicon-plus"/> Subscribe][N_SU] [<span class="glyphicon glyphicon-remove"/> Unsubscribe][N_UN] | |
+| **user**          | General user questions, help, and announcements  | [<i class="fa-solid fa-magnifying-glass"></i> Archive][U_A] | [<i class="fa-solid fa-plus"></i> Subscribe][U_SU] [<i class="fa-solid fa-xmark"></i> Unsubscribe][U_UN] | [<i class="fa-solid fa-envelope"></i> Post][U_P] |
+| **dev**           | Contributor discussions and development activity | [<i class="fa-solid fa-magnifying-glass"></i> Archive][D_A] | [<i class="fa-solid fa-plus"></i> Subscribe][D_SU] [<i class="fa-solid fa-xmark"></i> Unsubscribe][D_UN] | [<i class="fa-solid fa-envelope"></i> Post][D_P] |
```
ctubbsii, line 49: Why move away from the glyphicons halflings? Are those no longer allowed in newer bootstrap? Have you considered using [the bootstrap icons stuff](https://icons.getbootstrap.com/#install) rather than font-awesome?

pages/downloads.md
```diff
@@ -86,15 +86,14 @@ available in a future update to this site.
 {% assign lnkFile = 'accumulo-' | append: linkVers | append: '-' | append: srcbin | append: '.tar.gz' %}
 {% assign lnkSuffix = '/accumulo/' | append: linkVers | append: '/' | append: lnkFile %}
 <div class="row btn-group" style="margin-left: 20px; margin-bottom: 5px; display: block">
-  <div class="col btn-group">
+  <div class="col-auto btn-group">
     <a {{btnDownloadStyle}} href="{{closerLink}}{{lnkSuffix}}" link-suffix="{{lnkSuffix}}">{{lnkFile}}{{glyphSave}}</a>
-  </div><div class="col btn-group">
```
ctubbsii, line 91: You removed the separate div, but if I remember correctly, this was here to ensure the entire block wrapped as a group on screen width changes. Or maybe it was because it was for padding/spacing tweaks. I don't think it was extraneous, though. Same situation in the others below here.

pages/contact-us.md
```diff
@@ -45,10 +45,10 @@ follow discussions as they happen.
...
-| **dev**           | Contributor discussions and development activity | [<span class="glyphicon glyphicon-search"/> Archive][D_A] | [<span class="glyphicon glyphicon-plus"/> Subscribe][D_SU] [<span class="glyphicon glyphicon-remove"/> Unsubscribe][D_UN] | [<span class="glyphicon glyphicon-envelope"/> Post][D_P] |
-| **commits**       | Code changes                                     | [<span class="glyphicon glyphicon-search"/> Archive][C_A] | [<span class="glyphicon glyphicon-plus"/> Subscribe][C_SU] [<span class="glyphicon glyphicon-remove"/> Unsubscribe][C_UN] | |
-| **notifications** | Automated notifications (GitHub, etc.)             | [<span class="glyphicon glyphicon-search"/> Archive][N_A] | [<span class="glyphicon glyphicon-plus"/> Subscribe][N_SU] [<span class="glyphicon glyphicon-remove"/> Unsubscribe][N_UN] | |
+| **user**          | General user questions, help, and announcements  | [<i class="fa-solid fa-magnifying-glass"></i> Archive][U_A] | [<i class="fa-solid fa-plus"></i> Subscribe][U_SU] [<i class="fa-solid fa-xmark"></i> Unsubscribe][U_UN] | [<i class="fa-solid fa-envelope"></i> Post][U_P] |
```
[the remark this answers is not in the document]
DomGarguilo: `<i>` is for icons not italicizing afaik. Thats how its listed on the fontawesome icons docs: https://fontawesome.com/icons/plus?f=classic&s=solid

css/accumulo.scss
```diff
@@ -58,10 +58,6 @@ code {
     padding-left: 10px;
 }
```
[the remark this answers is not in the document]
DomGarguilo: Yea I've spotted a lot of places where custom css or inline `style`s could be replaced with bootstrap classes. I wanted to make minimal changes for the version upgrade and maybe come back in a separate PR to replace/clean up things where possible.

pages/contact-us.md
DomGarguilo: Yea glyphicons wasn't working. Switching to bootstrap icons would require an extra import but could be worth it potentially if we wanted to replace font-awesome with bootstrap icons but that should probably happen in a separate PR

pages/contact-us.md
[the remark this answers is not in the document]
ctubbsii: Nah, it's defined as "alternate voice or mood", typically presented as italics. There was an attempt with html4 to replace it with `<em>` to semantically mean "emphasis" and leave `<i>` to literally mean italics, but it never caught on because nobody ever styled `<em>` differently than `<i>`, and it was just a redundant tag. I think it was commonly used in html4 to represent an arbitrary inline element, because not all browsers consistently displayed `<span>` elements that were effectively empty. With html5, that's not really an issue anymore, because of the more widespread popularity of modern versions of CSS.

pages/contact-us.md
ctubbsii: Oh, I see. I didn't realize we were already using font-awesome stuff. That's fine, then. As long as we can use scalable fonts using the CSS classes, and don't have to deal directly with SVGs.

pages/contact-us.md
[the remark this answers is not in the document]
ctubbsii: I think one reason why people might prefer `<i>` is because it's shorter, and people seem to like minimizing. I don't find it helps that much, and I think it obfuscates things too much. Plus, if I actually want to put some text in there, like `<i class="fa-solid fa-magnifying-glass">something that wraps with the magnifying glass</i>`, then I have to specifically add CSS to *un*-style it.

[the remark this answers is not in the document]
DomGarguilo: I see. Thanks for explaining that. I can go ahead and change these to `<span>`. I was just going off of the fontawesome docs.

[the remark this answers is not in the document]
DomGarguilo: Resolved in 4bb6421

css/accumulo.scss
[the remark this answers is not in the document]
DomGarguilo: Created https://github.com/apache/accumulo-website/issues/401 to track this.

_includes/nav.html
```diff
@@ -1,64 +1,62 @@
...
+          </a>
           <ul class="dropdown-menu">
-            <li><a href="https://www.apache.org">Apache Homepage <i class="fa-solid fa-up-right-from-square"></i></a></li>
-            <li><a href="https://www.apache.org/licenses/">License <i class="fa-solid fa-up-right-from-square"></i></a></li>
-            <li><a href="https://www.apache.org/foundation/sponsorship">Sponsorship <i class="fa-solid fa-up-right-from-square"></i></a></li>
-            <li><a href="https://www.apache.org/security">Security <i class="fa-solid fa-up-right-from-square"></i></a></li>
-            <li><a href="https://www.apache.org/foundation/thanks">Thanks <i class="fa-solid fa-up-right-from-square"></i></a></li>
-            <li><a href="https://www.apache.org/foundation/policies/conduct">Code of Conduct <i class="fa-solid fa-up-right-from-square"></i></a></li>
-            <li><a href="https://www.apache.org/events/current-event.html">Current Event <i class="fa-solid fa-up-right-from-square"></i></a></li>
+            <li><a class="dropdown-item" href="https://www.apache.org">Apache Homepage <i class="fa-solid fa-up-right-from-square"></i></a></li>
```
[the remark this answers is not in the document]
DomGarguilo: Fixed in 6dd2577

pages/downloads.md
DomGarguilo: Fixed in b056d84

DomGarguilo:
Removing "WIP" as I think I've done everything that I can tell needs to be done for this upgrade.

The last thing was fixing the navbar which turns into a dropdown on small screens. This was fixed in 254cd06.

main:
boostrap3 collapse

with these changes:
boostrap5 collapse

_includes/nav.html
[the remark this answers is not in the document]
ctubbsii: I looked at the documentation, and I'm not sure I'm right about this. I framed it more as a question, because I wasn't sure. The documentation seems to indicate it's primarily for the keyboard navigation, not part of the visual of the menu. I'm still not sure why it's needed explicitly, instead of being inferred, but I don't have much free time to play with it.

I would follow the bootstrap examples, though. The correct answer to my previous questions probably should have been "no". Sorry if I created more churn by asking this.

pages/contact-us.md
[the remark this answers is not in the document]
ctubbsii: My main goal in asking about this was mostly to suggest not making unnecessary changes, in order to keep the diff minimal. As long as there's a reason to choose one over the other... whether it's "keep the diff minimal", "follow the bootstrap examples", "keep the code small", "consistency", or some other reason. I just might ask about those reasons, but that's not necessarily an objection.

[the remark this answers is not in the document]
DomGarguilo: > My main goal in asking about this was mostly to suggest not making unnecessary changes, in order to keep the diff minimal. As long as there's a reason to choose one over the other... whether it's "keep the diff minimal", "follow the bootstrap examples", "keep the code small", "consistency", or some other reason. I just might ask about those reasons, but that's not necessarily an objection.

I didn't have much of an opinion one way or another and you made some good points about using span instead of `<i>` so I think they are fine to stay as `<span>`

_includes/nav.html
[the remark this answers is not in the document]
DomGarguilo: I'll take a second look and change things back if needed. Everything seems to be working okay as is though.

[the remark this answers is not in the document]
DomGarguilo: Yea it looks like convention is to have the class within the `<a>` so I moved it back in 239694b

pages/downloads.md
```diff
@@ -12,8 +12,8 @@ var updateLinks = function(mirror) {
 };
 
 var mirrorsCallback = function(json) {
-  var htmlContent = '<div class="row"><div class="col-md-3"><h5>Select an Apache download mirror:</h5></div>' +
-    '<div class="col-md-5"><select class="form-control" id="apache-mirror-select">';
+  var htmlContent = '<div class="row align-items-center mb-3"><div class="col-3"><h5>Select an Apache download mirror:</h5></div>' +
+    '<div class="col-5"><select class="form-select" id="apache-mirror-select">';
```
ctubbsii, line 16:
```suggestion
    '<div class="col-5"><select class="form-select" id="apache-mirror-select">';
```

pages/downloads.md
```diff
@@ -85,26 +85,29 @@ available in a future update to this site.
...
+<div class="row" style="margin-left: 20px; margin-bottom: 5px;">
+  <div class="btn-group col-auto">
     <a {{btnDownloadStyle}} href="{{closerLink}}{{lnkSuffix}}" link-suffix="{{lnkSuffix}}">{{lnkFile}}{{glyphSave}}</a>
-  </div><div class="col btn-group">
-    <a {{btnSigStyle}} href="{{downloadsLink}}{{lnkSuffix}}.asc">ASC{{glyphLock}}</a>
-    <a {{btnHashStyle}} href="{{downloadsLink}}{{lnkSuffix}}.sha512">SHA{{glyphLock}}</a>
+    <div class="btn-group">
+      <a {{btnSigStyle}} href="{{downloadsLink}}{{lnkSuffix}}.asc">ASC{{glyphLock}}</a>
+      <a {{btnHashStyle}} href="{{downloadsLink}}{{lnkSuffix}}.sha512">SHA{{glyphLock}}</a>
+    </div>
```
ctubbsii, line 94: This still looks substantially different from what it was before, and I'm not sure if it has the same behavior as before. Originally, it was two button groups adjacent to one another, as if two columns in the same row. Here, you've changed it to a button group nested inside another, and the row effectively has one column now (which obviates the value of having a row).

I think it should be:

```suggestion
  </div><div class="btn-group">
      <a {{btnSigStyle}} href="{{downloadsLink}}{{lnkSuffix}}.asc">ASC{{glyphLock}}</a>
      <a {{btnHashStyle}} href="{{downloadsLink}}{{lnkSuffix}}.sha512">SHA{{glyphLock}}</a>
```

Where this matters is when you make the window size very small. This makes the buttons wrap in a very specific way that keeps the validation buttons (ASC, SHA) together, and situated just beneath the link to the file (rather than forcing the user to scroll), but without allowing them to easily wrap, creating a column of buttons mangled together with no logical groupings at all.

pages/downloads.md
```diff
@@ -85,26 +85,29 @@ available in a future update to this site.
...
 </div>
 {% endfor %}
-<div class="row btn-group-sm" style="margin: 20px">
-  <a {{btnDocStyle}} href="{{site.baseurl}}/release/accumulo-{{linkVers}}">Release Notes</a>
-  <a {{btnDocStyle}} href="https://github.com/apache/accumulo/blob/rel/{{linkVers}}/README.md">README</a>
-  <a {{btnDocStyle}} href="{{site.baseurl}}/docs/2.x">Online Documentation</a>
-  <a {{btnDocStyle}} href="https://github.com/apache/accumulo-examples">Examples</a>
-  <a {{btnDocStyle}} href="{{site.baseurl}}/docs/2.x/apidocs3">Java API</a>
+<div class="row" style="margin: 20px;">
+  <div class="btn-group-sm">
```
ctubbsii, line 99: This extra div creates a single column inside the row div. Previously, the intent was to have each `<a>` element be a separate column within the row.

ctubbsii (changes requested):
Overall, looks good, but I would suggest trying to preserve the previous structure of the downloads page buttons, because I think the changes here will cause the buttons to wrap more poorly than they currently do. I remember spending a lot of effort to get them to wrap just right, so everything looked good on mobile (or just small screens) and on a regular sized screen, and I think some of the structural changes to the div tags breaks some of that.

Everything else on the site seems good, though.

pages/downloads.md
DomGarguilo: Should be fixed as of 531aded.

Here is a screencap with only the top one fixed to show the difference:
new button wrap

And here is a screencap with all fixed (how it appears with the commit above):
whole page wrap 5

pages/downloads.md
DomGarguilo: In a713543 I removed the unneeded `row` div that wrapped the `btn-group-sm` div and it seems to work well.

pages/downloads.md
DomGarguilo: addressed in a713543

pages/downloads.md
```diff
@@ -85,16 +85,17 @@ available in a future update to this site.
...
 {% assign lnkFile = 'accumulo-' | append: linkVers | append: '-' | append: srcbin | append: '.tar.gz' %}
 {% assign lnkSuffix = '/accumulo/' | append: linkVers | append: '/' | append: lnkFile %}
-<div class="row btn-group" style="margin-left: 20px; margin-bottom: 5px; display: block">
-  <div class="col btn-group">
+<div class="d-flex flex-wrap justify-content-start align-items-start" style="margin-left: 20px; margin-bottom: 5px;">
+  <div class="btn-group">
     <a {{btnDownloadStyle}} href="{{closerLink}}{{lnkSuffix}}" link-suffix="{{lnkSuffix}}">{{lnkFile}}{{glyphSave}}</a>
-  </div><div class="col btn-group">
+  </div>
+  <div class="btn-group">
```
ctubbsii, line 92: I'm not familiar with any of these classes and how they affect wrapping, but I'd be careful about introducing whitespace between the two divs. The old one had them directly adjacent, with no whitespace in-between, so the padding and wrapping worked correctly. Changing it is probably fine, so long as you've tested it. If it gets quirky, we can always revisit it and fix it later as a follow-on.

pages/downloads.md
ctubbsii: Yeah, that change should be fine. The div with the `btn-group-sm` is already acting as the "row" container anyway.

pages/downloads.md
ctubbsii: The old buttons were lighter, and had mouse-over effects that made them slightly darker as a highlight before you click them. I'm not sure if the new ones have that... but one thing that is clearly different is that the ASC and SHA buttons seem to blend together. In the old version, they had a clear border, so you could tell they were separate buttons. We may need to fix the padding for those.

ctubbsii: One other thing I noticed is the lock icons probably need a `&nbsp;` between ASC and the icon and between SHA and the icon, to prevent the icon from wrapping on its own. It should stay with the text.

ctubbsii: Oh, I found the colors do have a lighter variant that look a bit better. They are the `-subtle` versions. I don't see subtle options for buttons, only for badges. I thought maybe you could get away with just using `btn bg-primary-subtle`, for example, instead of `btn btn-primary`, but the badges don't have the hover action like buttons. So, we'd just have to override the colors for the button classes in our own CSS to the same as what the subtle versions are, if we want to use the lighter colors and still have a mouseover effect that tints them slightly darker.

DomGarguilo: > The old buttons were lighter, and had mouse-over effects that made them slightly darker as a highlight before you click them.

Here is how the new buttons behave:
button hover

> the ASC and SHA buttons seem to blend together. In the old version, they had a clear border

I noticed this as well but that seems to be the new bootstrap default. There are classes like `btn-outline-secondary` that have a light background and the outline and text is of the secondary color (see example below) but not sure if thats something we want to do. We could also override the button styles as you suggested in the other comment.
image

> the lock icons probably need a &nbsp; between ASC and the icon and between SHA and the icon

This is how the code is now in this branch. `{% assign glyphSave = '&nbsp;<span class="fa-solid fa-cloud-arrow-down"></span>' %}`

ctubbsii: Oh, I like the outline variant. Let's go with that one. Much easier to not have to mess with colors. I still think we need to add some padding between the button groups, though. They're a bit mashed up against one another. The spacing between the bottom row (release notes, readme, etc.) are nice. It'd be nice to have the same between the download links and the checksum buttons. The line between the ASC/SHA buttons are sufficient, though, so we don't need a space there... it seems the style as a "button group" there is fine (whereas I was previously thinking it as a "group of buttons" rather than a "button group").

ctubbsii: > This is how the code is now in this branch. `{% assign glyphSave = '&nbsp;<span class="fa-solid fa-cloud-arrow-down"></span>' %}`

That's for the arrow for the download link. I assume there's a similar for the ASC/SHA, but it didn't look like it worked in your visualization. Not sure why.

DomGarguilo: > > This is how the code is now in this branch. `{% assign glyphSave = '&nbsp;<span class="fa-solid fa-cloud-arrow-down"></span>' %}`
> 
> That's for the arrow for the download link. I assume there's a similar for the ASC/SHA, but it didn't look like it worked in your visualization. Not sure why.

`{% assign glyphLock = '&nbsp;<span class="fa-solid fa-lock"></span>' %}`. The lock icon code has it too. Which visualization doesn't look right?

DomGarguilo: In bad8a2d I added a slight margin to the right of the download button and made the other two buttons the outline style:
Screenshot from 2023-09-12 14-33-21
And when wrapping:
Screenshot from 2023-09-12 14-33-35

Alternatively, here is what it would look like for the download button to have the outline style too:
Screenshot from 2023-09-12 14-34-19

ctubbsii (approved):

PR #400, opened 2023-08-29
Use bootstrap styling for dataTables

ctubbsii:
* Use Bootstrap 3 instead of dataTables default styling, so the tables can fit the rest of the site's theme by default
* This will need to be switched to Bootstrap 5 versions when Bootstrap is updated

Follow up from #396


ctubbsii:
> LGTM. The styling before and after looks identical as far as I can tell.

There are a few extra stray borders and there are some slight padding differences, due to the competing styles. It would be more obvious if we customized the bootstrap table theme. I'm not suggesting we do that, but this should ensure it stays consistent with the rest of the site, if we happen to do that, and it resolves those few minor competing style issues. Thanks for merging.

PR #405, opened 2023-09-08
Mark the link to our Twitter presence as inactive

ctubbsii:
More prominently indicate that our use of Twitter is inactive, due to the inability of the PMC to properly share the responsibilities of maintaining our presence there (TweetDeck is now a paid premium service, and we cannot use it any longer for free) and the lack of interest from the PMC in maintaining our presence there.


DomGarguilo (approved):
LTGM

Here is a screenshot with these changes:
image

ctubbsii:
Oh, I guess I didn't realize this got rendered to the main page. I was thinking it was only on a separate page of links or on the contact-us page. For the main page, I think it should probably just be omitted.

Also, I'm not sure why LinkedIn is so prominent on our main page. It's only tangentially related to the project, and as far as I am aware, the group there is not maintained by the PMC. That should probably be a footnote on the people page rather than so prominent on the main page. It certainly shouldn't be conveyed higher in the list than the GitHub link to the repo, which is actually a PMC-maintained space.

I think I'm going to take another pass at this.

ctubbsii:
Updated title and description to match these new changes... this time, I got rid of the box entirely from the page. It's not really needed there.

DomGarguilo (commented):
Sorry didnt get a change to re-review before this was merged but everything LGTM

PR #410, opened 2023-12-18
Bump ruby version in github actions script

DomGarguilo:
It looks like the build is failing with:
```
  bundle exec jekyll build
  shell: /usr/bin/bash -e {0}
  ruby 2.6.10p210 (2022-04-12 revision 67958) [x86_64-linux]
  ERROR:  Error installing bundler:
	  The last version of bundler (>= 0) to support your Ruby & RubyGems was 2.4.22. Try installing it with `gem install bundler -v 2.4.22`
	  bundler requires Ruby version >= 3.0.0. The current ruby version is 2.6.10.210.
```
This PR moves to the version found in the Dockerfile, 3.2.2

### DIFF
--- a/.github/workflows/jekyll.yaml
+++ b/.github/workflows/jekyll.yaml
@@ -34,7 +34,7 @@ jobs:
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 2.6
+        ruby-version: 3.2.2
     - name: Cache gems
       uses: actions/cache@v3
       with:


DomGarguilo:
The build passed so I am going to assume this change is okay. I'm going to merge it in now since there are other changes waiting for the build to pass but feel free to change what I have if it doesn't look right.

ctubbsii:
Just for reference, ASF Infrastructure, as far as I know, might still be using 2.5.1 (https://github.com/apache/accumulo-website/pull/389#issuecomment-1654726554). The further we move away from that in our testing, the less likely it will still work in ASF's infrastructure. So, we just need to be careful to watch any consequences of bumping this.

PR #411, opened 2023-12-19
Update documentation to create an empty wal file

EdColeman:
Updated the documentation for creating an empty wal file to use version 4 header, a uuid and other information.

### DIFF
--- a/_docs-2/troubleshooting/advanced.md
+++ b/_docs-2/troubleshooting/advanced.md
@@ -249,21 +249,29 @@ metadata table!), the following process can be followed to create a valid, empty
 WAL file. Run the following commands as the Accumulo unix user (to ensure that
 the proper file permissions in HDFS)
 
-    $ echo -n -e '--- Log File Header (v2) ---\x00\x00\x00\x00' > empty.wal
+```
+$ UUID=$(uuidgen); \
+echo -n -e '--- Log File Header (v4) ---U+1F47B$'"${UUID}"'\x00\x00\x00\x00' > "${UUID}".wal; \
+echo 'created: '"${UUID}"'.wal'
+```
 
-The above creates a file with the text "--- Log File Header (v2) ---" and then
-four bytes. You should verify the contents of the file with a hexdump tool.
+The above creates a file with the text "--- Log File Header (v4) ---" a unicode character to flag no decryption 
+parameters, a UUID and then four bytes. The file created will be `[uuid]`.wal and the name is echoed to the command
+line. You should verify the contents of the file with a hexdump tool.
 
-Then, place this empty WAL in HDFS and then replace the corrupt WAL file in HDFS
-with the empty WAL.
+Then, place this empty WAL in HDFS and then replace the corrupt WAL file in HDFS with the empty WAL for the 
+tserver / host pair with the following hdfs commands:
 
-    $ hdfs dfs -moveFromLocal empty.wal /user/accumulo/empty.wal
-    $ hdfs dfs -mv /user/accumulo/empty.wal /accumulo/wal/tserver-4.example.com+10011/26abec5b-63e7-40dd-9fa1-b8ad2436606e
+    $ hdfs dfs -moveFromLocal [uuid].wal /user/accumulo/[uuid].wal
+    $ hdfs dfs -mv /user/accumulo/[uuid].wal /accumulo/wal/[tserver+port]/[uuid]
 
-After the corrupt WAL file has been replaced, the system should automatically recover.
-It may be necessary to restart the Accumulo Manager process as an exponential
-backup policy is used which could lead to a long wait before Accumulo will
-try to re-load the WAL file.
+Note:
+- the `+` separator for port.
+- the wal file is the uuid, without an extension
+
+After the corrupt WAL file has been replaced, the system should automatically recover. It may be necessary to restart 
+the Accumulo Manager process as an exponential backup policy is used which could lead to a long wait before Accumulo 
+will try to re-load the WAL file.
 
 ## Zookeeper Failures
 


_docs-2/troubleshooting/advanced.md
```diff
@@ -249,21 +249,29 @@ metadata table!), the following process can be followed to create a valid, empty
...
 
-Then, place this empty WAL in HDFS and then replace the corrupt WAL file in HDFS
-with the empty WAL.
+Then, place this empty WAL in HDFS and then replace the corrupt WAL file in HDFS with the empty WAL for the 
+tserver / host pair with the following hdfs commands:
 
-    $ hdfs dfs -moveFromLocal empty.wal /user/accumulo/empty.wal
-    $ hdfs dfs -mv /user/accumulo/empty.wal /accumulo/wal/tserver-4.example.com+10011/26abec5b-63e7-40dd-9fa1-b8ad2436606e
+    $ hdfs dfs -moveFromLocal [uuid].wal /user/accumulo/[uuid].wal
+    $ hdfs dfs -mv /user/accumulo/[uuid].wal /accumulo/wal/[tserver+port]/[uuid]
```
ctubbsii, line 266:
````suggestion
```sh
hdfs dfs -moveFromLocal "$UUID.wal" "/user/accumulo/$UUID.wal"
hdfs dfs -mv "/user/accumulo/$UUID.wal" "/accumulo/wal/[tserver+port]/$UUID"
```
````

File: _docs-2/troubleshooting/advanced.md [hunk 2]
```diff
@@ -249,21 +249,29 @@ metadata table!), the following process can be followed to create a valid, empty
...
-After the corrupt WAL file has been replaced, the system should automatically recover.
-It may be necessary to restart the Accumulo Manager process as an exponential
-backup policy is used which could lead to a long wait before Accumulo will
-try to re-load the WAL file.
+Note:
+- the `+` separator for port.
+- the wal file is the uuid, without an extension
+
+After the corrupt WAL file has been replaced, the system should automatically recover. It may be necessary to restart 
+the Accumulo Manager process as an exponential backup policy is used which could lead to a long wait before Accumulo
```
ctubbsii, line 273:
```suggestion
the Accumulo Manager process to reduce the recovery time, because an exponential backup policy is used which could lead to a long wait before Accumulo 
```

_docs-2/troubleshooting/advanced.md
```diff
@@ -249,21 +249,29 @@ metadata table!), the following process can be followed to create a valid, empty
...
-It may be necessary to restart the Accumulo Manager process as an exponential
-backup policy is used which could lead to a long wait before Accumulo will
-try to re-load the WAL file.
+Note:
+- the `+` separator for port.
+- the wal file is the uuid, without an extension
+
+After the corrupt WAL file has been replaced, the system should automatically recover. It may be necessary to restart 
+the Accumulo Manager process as an exponential backup policy is used which could lead to a long wait before Accumulo 
+will try to re-load the WAL file.
```
ctubbsii, line 274:
```suggestion
will try to reload the WAL file.
```

ctubbsii (changes requested):

ctubbsii (approved):

_docs-2/troubleshooting/advanced.md
```diff
@@ -249,21 +249,31 @@ metadata table!), the following process can be followed to create a valid, empty
...
-
-    $ hdfs dfs -moveFromLocal empty.wal /user/accumulo/empty.wal
-    $ hdfs dfs -mv /user/accumulo/empty.wal /accumulo/wal/tserver-4.example.com+10011/26abec5b-63e7-40dd-9fa1-b8ad2436606e
-
-After the corrupt WAL file has been replaced, the system should automatically recover.
-It may be necessary to restart the Accumulo Manager process as an exponential
-backup policy is used which could lead to a long wait before Accumulo will
-try to re-load the WAL file.
+```sh
+UUID=$(uuidgen); echo -n -e '--- Log File Header (v4) ---U+1F47B$'"$UUID"'\x00\x00\x00\x00' >"$UUID".wal
```
ctubbsii, line 253: So, I was tracking down where the `U+1F47B` comes from. I think the original intent is to have the actual ghost emoji character stored there, but instead, it's a human-readable representation of the unicode code point for the ghost emoji that's stored there. It doesn't really matter what it's value is, though. It's purpose is to be a fixed value that represents the decryption parameters for a "not encrypted" file.

Specifically, this value is an indicator that it's "not encrypted" using our built-in non-encrypting CryptoService that we use in our built-in reference implementations for a CryptoServiceFactory. A user could provide their own factory that supports a "not encrypted" option using a different token to identify files it supports. But, this string is to denote ours.

I checked the serialization of our WALs in the code, and it looks like this is wrong as written. There should be an integer that is written before the "U+1F47B" part. After that, I'm not sure what the WAL is supposed to contain, because I only got so far tracing the code to see how it writes the WAL header.

In Java, this is what I saw as a header, when tracing the steps to write out the size of the crypto params, then writing out the string for the "no crypto" params:

```java
byte[] header = "--- Log File Header (v4) ---\000\000\000\007U+1F47B".getBytes(UTF_8)
// { 45, 45, 45, 32, 76, 111, 103, 32, 70, 105, 108, 101, 32, 72, 101, 97, 100, 101, 114, 32, 40, 118, 52, 41, 32, 45, 45, 45, 0, 0, 0, 7, 85, 43, 49, 70, 52, 55, 66 }
```

That's equivalent in bash to:

```sh
UUID=$(uuidgen); echo -n -e '--- Log File Header (v4) ---\x00\x00\x00\x07U+1F47B' >"$UUID".wal
```

I don't know where the UUID comes in, or that dollar sign in the above example, or the zeroes from the v2 format. There may be more needed after this header to result in a valid WAL file... I'm not sure, but I do think what's here now in this PR is wrong.

ctubbsii (changes requested):

_docs-2/troubleshooting/advanced.md
EdColeman: As discussed, I'm creating a utility similar to the one to create an empty rfile. When complete these instructions will be updated.

EdColeman: The documentation has been updated to remove the previous changes and add the create-empty WAL / RFILE options.

PR #415, opened 2024-02-20
Adding Accumulo-Access to web site

dlmarion:


_includes/nav.html
```diff
@@ -42,6 +42,7 @@
             <li><a class="dropdown-item" href="{{ site.baseurl }}/related-projects">Related Projects</a></li>
           </ul>
         </li>
+        <li class="nav-item"><a class="nav-link" href="{{ site.baseurl }}/accumulo-access">Accumulo Access</a></li>
```
DomGarguilo, line 45: Kinda seems out of place to add a new item to the nav bar just for accumulo access. It could be cool to have a section to link to all of the adjacent accumulo projects/repos with a brief description of what each is (like accumulo-testing, access, docker, etc.). Not sure where the best place for that is though and also might not be within the scope of this PR.

Just an idea too. Not recommending we add it, just wanted to get the idea out there.

keith-turner: Since there is a voted on release we could add accumulo access to the following page instead of adding a new menu item.  None of the other git repos have a voted on release.  The download page could have a brief description and link to the GH readme in addition to a link to the src tarball.

https://accumulo.apache.org/downloads/

dlmarion: Both are good suggestions - my approach was to create something simple that may not need to be updated in the future by just providing links to the appropriate pages. I'm fine with whatever TBH, but I'll leave this open to continue the discussion so that we can come to some resolution.

ctubbsii: > None of the other git repos have a voted on release.

The accumulo2-maven-plugin does have one, but it's not really useful as a download on its own. But I agree maybe there should be something on the downloads page.

dlmarion:
Closing in favor of #443

PR #418, opened 2024-02-23
Fix broken link

jmark99:
When clicking on the `Contribute` link for the Core Project under the Accumulo Repositories list, the link presents a 404 error. This PR updates the link to point to the proper page.


ctubbsii (approved):
I think any other repos should be similarly changed to put theirs in their .github directories as well.

PR #422, opened 2024-04-10
Small improvements to development dockerfile

DomGarguilo:
The following changes were made:
* `--force-polling` was removed. It was causing one thread at a time on my machine to jump to 100% CPU usage. Removing it seems to have no negative affects.
* `--incremental` was added as an optimization. It makes it so it only regenerates the files that have changed instead of the whole project.
* Added more directories and files to the `exclude` field in the _config.yml so that when those files change, the site is not regenerated. Lots of the dirs and files I borrowed from the .gitignore


Dockerfile
```diff
@@ -31,4 +31,4 @@ ENV PORT=4000
 EXPOSE $PORT
 
 # Configure the default command to build from the mounted repository.
-CMD bundle exec jekyll serve --force-polling -H $HOST -P $PORT
+CMD bundle exec jekyll serve --incremental -H $HOST -P $PORT
```
ctubbsii, line 34: Incremental is still marked as experimental, and the docs say it may break the site in some cases. If it's substantially more efficient, then it's fine if you leave it, but if it's only marginally beneficial, I'd drop it.

ddanielr (approved):

PR #423, opened 2024-05-01
Update site to comply with ASF privacy site checks

### LINKED ISSUE #420: Resolve ASF web-checker issues.
The ASF has a tool that examines the web-site and reports potential issues.  The report can be seen at [Accumulo Site Check](https://whimsy.apache.org/site/project/accumulo)

Currently there are two issues flagged.  Missing link to ASF privacy policy and links to external resources.

According to the tool, the privacy policy is expected to be 
```
\Ahttps://privacy\.apache\.org/policies/privacy-policy-public\.html\z | \Ahttps?://(?:www\.)?apache\.org/foundation/policies/privacy\.html\z
```
The reported external resources are:
```
Found 6 external resources: {"cdn.jsdelivr.net"=>2, "cdnjs.cloudflare.com"=>2, "cdn.datatables.net"=>2}
```

ctubbsii:
See https://whimsy.apache.org/site/project/accumulo

* Add link to ASF privacy policy in ASF links section of navigation
* Stop linking to external CDNs for our default page layout. Instead, copy the resources, so they are hosted by our site. We don't really need the integrity checks to verify the downloads from CDN sites anymore. It adds no security since we're hosting them on the same service that would provide the checksum (same security domain for users to trust), and removing them alleviates the browser from that unnecessary work.
* Standardize how resources are specified, so they are more consistent and easier to update, including converting the embedded `<script>` from `_includes/scripts.html`, which contained our own custom scripts for our site, into its own `.js` file.
* Include free web fonts from fontawesome project, so the css can link
  to the ones we're hosting (downloaded from
  https://fontawesome.com/download, specifically
  https://use.fontawesome.com/releases/v6.4.2/fontawesome-free-6.4.2-web.zip)

This fixes #420


ctubbsii:
I don't think we should do this, as explained above, but if one really wanted to add the integrity checks back in, it would be something like:

```bash
shopt -s globstar
algo=sha384 # could replace with sha512
for x in css/*/**/*.css; do
  echo "<link rel=\"stylesheet\" type=\"text/css\" integrity=\"$algo-$(<$x openssl dgst -binary -$algo|openssl base64 -A)\" href=\"{{ site.baseurl }}/$x\">"
done
for x in js/*/**/*.js; do
  echo "<script type=\"text/javascript\" integrity=\"$algo-$(<$x openssl dgst -binary -$algo|openssl base64 -A)\" src=\"{{ site.baseurl }}/$x\"></script>"
done
```
(Note: I'm not sure if the order matters... some resources depend on other resources. So, you might not be able to use this output directly, but may have to re-order things.)

EdColeman (approved):

EdColeman (approved):

ctubbsii:
> Wasn't there a discussion somewhere else about not using minified resources? Or does that not apply in this case?

That is true for project source code that we release. Minified resources are not considered "open source" software, so we should not distribute them as such. However, for the website, we're just using them as libraries... we're not distributing them as part of our open source software releases.

PR #424, opened 2024-05-01
Update minimum Ruby version and gems

ctubbsii:
* Update minimum Ruby version to at least the version run by ASF infra team on Jekyll builders, so that way we can reduce compatibility surprises when trying to build on stuff substantially newer than what's on ASF's infrastructure
* Do a bundle update


ctubbsii:
@clambertus confirmed in Slack that the builder is running Ruby 2.7 (I think it's safe to rely on it not going backwards). These updated gems work fine on Ruby 3.3.0 on my Fedora 40 dev machine. I imagine they work fine with the 3.2.2 version used by our Docker container for testing locally, but I can't test that, since I don't have Docker set up locally. Maybe @ddanielr or @DomGarguilo has that set up already and can test it?

DomGarguilo (approved):
LGTM - Works for me when using the development docker container

EdColeman (approved):
The changes look okay - I'll defer to @DomGarguilo for running it in docker.

PR #426, opened 2024-05-02
Add bootstrap map files

### LINKED ISSUE #425: Error when running development docker container - map files not found
When I run the development docker container, I am seeing new errors in the server logs when I visit the webpage:
```
ERROR `/js/bootstrap/5.3.1/dist/js/bootstrap.bundle.min.js.map' not found.
ERROR `/css/bootstrap/5.3.1/dist/css/bootstrap.min.css.map' not found.
```

I think this is happening after the changes in #423 where the resources are now served from our site as opposed from a cdn. I don't see a reason why we shouldn't just include these map files along side the other resources but there might be a reason I am not thinking of.

DomGarguilo:
Fixes #425 

Adds the map files for the minified boostrap resources added in #423

These map files were download from https://github.com/twbs/bootstrap/archive/v5.3.1.zip


ctubbsii (changes requested):
I made a comment at https://github.com/apache/accumulo-website/issues/425#issuecomment-2137784373 that suggested we just use the non-minified versions instead of bothering with these .map files.

DomGarguilo:
In a27b55f I replaced the `.min` and `.map` files with the single non-minified versions. I also took this opportunity to bump to bootstrap version 5.3.3

ctubbsii (approved):
These changes look good to me. Two questions I have are:

* Does this resolve the issue you were seeing?
* Is it worth changing the other minified resources to their regular versions also?

DomGarguilo:
> These changes look good to me. Two questions I have are:
> 
> * Does this resolve the issue you were seeing?

Yes, the error I was seeing is no longer present.

> * Is it worth changing the other minified resources to their regular versions also?

It is an easy enough change to make so I could convert the others too as part of a version bump in a separate PR.

DomGarguilo:
> > These changes look good to me. Two questions I have are:
> > 
> > * Does this resolve the issue you were seeing?
> 
> Yes, the error I was seeing is no longer present.

I guess I spoke too soon. After a while of the development docker container running, I still see the original errors (except now they mention the non minified versions):
```
[2024-07-17 23:48:45] ERROR `/css/bootstrap/5.3.3/dist/css/bootstrap.css.map' not found.
[2024-07-17 23:48:45] ERROR `/js/bootstrap/5.3.3/dist/js/bootstrap.bundle.js.map' not found.
```

ctubbsii:
Hmm, that's very weird. I'd like to try to reproduce this. I have no idea what is creating these errors. The files themselves are definitely not needed.

DomGarguilo:
> Hmm, that's very weird. I'd like to try to reproduce this. I have no idea what is creating these errors. The files themselves are definitely not needed.

I am running things in the development docker container. Should pop up there. The files are refferenced at the bottom of the files. For example:
https://github.com/apache/accumulo-website/blob/1b64cd852624e0ebb47fb15d6da315d4b911292b/js/bootstrap/5.3.3/dist/js/bootstrap.bundle.js#L6314

It has the same for the non minified too.

ctubbsii:
> > Hmm, that's very weird. I'd like to try to reproduce this. I have no idea what is creating these errors. The files themselves are definitely not needed.
> 
> I am running things in the development docker container. Should pop up there. The files are refferenced at the bottom of the files. For example:

The Dockerfile just seems to run `bundle exec jekyll serve`. I don't get any errors when I run that locally, so I'm very curious what's different. Perhaps it's a different version of jekyll, or some dependency? I'm not sure where the errors are appearing, or which command they are coming from. I don't have Docker set up, so I can't reproduce this.

I did find https://stackoverflow.com/questions/57582243/no-sourcemap-for-css
This might be some difference in default behavior coming from different versions of jekyll, or sass, or something. Perhaps we can force one behavior or another by setting configuration?

ctubbsii:
Also, perhaps we can just ignore the errors? They seem to be false positives of some sort anyway. I don't think they represent an actual error in anything.

DomGarguilo:
> Also, perhaps we can just ignore the errors? They seem to be false positives of some sort anyway. I don't think they represent an actual error in anything.

Yea it should be fine to ignore them. Especially if they are only appearing in the development docker output.

PR #428, opened 2024-07-17
Replace minified resources with non-minified. Update versions

DomGarguilo:
Bump jquery from 3.7.0 to 3.7.1. replace with non-minified version
Bump fontawesome from 6.4.2 to 6.6.0. Use non-minified css file
Bump datatables from 1.13.6 to 2.0.8. Use non-minified css and js files

These changes were originally discussed [in this thread](https://github.com/apache/accumulo-website/pull/426#pullrequestreview-2175517135)


ctubbsii (approved):
The changes look fine to me. But, am I correct in saying that you still see the errors in the Docker container about the missing map files even after these changes?

PR #431, opened 2024-07-23
Proposed update to bylaws

ctubbsii:
Simplify the bylaws by removing extraneous details about the ASF and its standard practices, instead of embedding everything into the bylaws. Also align them to how we actually operate, removing sections that are either wrong or unused.

Notable changes are:

* Bump bylaws to version 4 (if approved)
* Add link to The Apache Way briefing
* Simplified introduction to the bylaws themselves
* Remove redundant role definitions and link to Foundation descriptions
* Align description about how PMC members are added with the board procedures, and write it in a way that preserves our normal voting practice for new members while not undermining the board procedures or subtracting from the PMC Chair's delegated authority
* Align the description of Emeritus PMC member privileges with one's actual privileges granted as a position within the Foundation (for example, remove the part about keeping voting rights... that's not a thing upon resigning from the PMC). However, make it clear that we still allow one to stay on the private list if they are Emeritus (we can decide to do things differently and remove those people... but I wasn't going to propose that change here)
* Reorder PMC Chair description after PMC member section and clarify expectations to make a good faith effort to get consensus from the PMC members
* Drop everything about formal release plans. We don't ever actually construct them formally... the release process is too simple and automated to require formal release plans, and this whole section just creates unnecessary bureaucracy.
* However, do add extra information about what kinds of things a release manager is expected to do as part of curating a release
* Drop all the descriptions of voting types and link to glossary
* Drop descriptions about how voting happens, and link to the Foundation page on voting
* Add details for how a vote subject line is formatted, how the result message is formatted, and how votes are closed in our project
* Drop all the details about the different circumstances where we vote and what vote type we use, and simplify it to how we actually work, which is essentially consensus approval of at least 3 days for everything, lazy consensus on smaller matters, and majority approval for releases. The vote circumstances this drops are the release plan creation stuff that we don't use, and the adopting new code base... which isn't really a thing we've had to deal with and doesn't really require a special line item, as it's just a regular consensus approval, which is our basic vote type. This does drop all the vote durations to 3 days... if we really want to preserve the 7 days for specific circumstances, that can be added back in before this proposal is adopted, but I think 3 days is generally enough.

In addition to the proposed changes to the bylaws, this change also includes dropping two pages from our docs that are redundant, describing voting in general and verbosely explaining lazy consensus. The Foundation pages, glossary, and the bylaws themselves are sufficient for these, and don't require separate verbose pages to explain.


contributor/bylaws.md
```diff
@@ -154,162 +140,88 @@ An Accumulo release manager is also expected to:
...
+
+For more information on how to vote, see the Foundation's page on
+[voting][voting].
+
+The Foundation defines voting criteria for procedural issues, code
+modifications, and releases. Most formal votes are going to be [consensus
+approval][consensus]. Release votes, however, follow [majority
+approval][majority]. Other decisions, when necessary, can often be made through
+[lazy consensus][lazy]. In the case of an objection for a lazy consensus vote,
+or the desire for explicit consensus, one can initiate a formal vote thread.
```
keith-turner, line 187:
```suggestion
The Foundation defines voting criteria for procedural issues, code
modifications, and releases. Most formal votes will be [consensus
approval][consensus]. Release votes, however, follow [majority
approval][majority]. Other decisions, when necessary, can often be made through
[lazy consensus][lazy]. In the case of an objection for a lazy consensus vote,
or the desire for explicit consensus, one can initiate a formal vote thread.
```

contributor/bylaws.md
```diff
@@ -154,162 +140,88 @@ An Accumulo release manager is also expected to:
...
-After feature freeze, new features should not be accepted for the release.
-After code freeze, only critical fixes should be accepted for the release. The
-release manager guides the decision on accepting changes.
-
-All dates in a plan are estimates, as unforeseen issues may require delays. The
-release manager may adjust dates as needed. In serious circumstances, the
-release manager may opt to call a re-plan vote.
+Accumulo follows a commit-then-review (CTR) policy. This only means that it is
+not strictly a requirement to achieve consensus prior to committing code
+changes to the code repository. Committers can make changes to the codebase
```
keith-turner, line 195:
```suggestion
Accumulo follows a commit-then-review (CTR) policy. This means that consensus
is not required prior to committing. Committers can make changes to the codebase
```

keith-turner (approved):
These changes read really nicely.  Made some suggestions about the wording of things, feel free to ignore those.  I was just attempting to make some things more concise, but there was no problem with the original.

contributor/bylaws.md
```diff
@@ -154,162 +140,88 @@ An Accumulo release manager is also expected to:
...
-votes of active PMC members are binding.
-
-See the [voting page][voting] for more details on the mechanics of voting.
+starting with `[VOTE]`. After the vote period has elapsed, the initiator of the
+vote, or their designee, closes it by replying to the thread with the results
+of the vote. That result email should use the same subject line preceded by
+`[RESULT][VOTE]`. Voting is carried out by replying to the vote mail and
+continues until the vote is closed. If a vote thread becomes inactive and
+remains open for too long, without a response from the initiator, the PMC Chair
+may close the vote.
```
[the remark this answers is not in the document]
keith-turner: Thought I was going to change more and then decided to only change a little bit.  That is why there is a big suggestion block w/ a small change.

cshannon (approved):
I read through the changes and they seem fine to me.

brianloss (commented):
Thanks for doing this @ctubbsii. The changes look good to me--just had the one question.

contributor/bylaws.md
```diff
@@ -154,162 +140,87 @@ An Accumulo release manager is also expected to:
...
-an action across the wider Accumulo community. For PMC decisions, only the
-votes of active PMC members are binding.
-
-See the [voting page][voting] for more details on the mechanics of voting.
+starting with `[VOTE]`. After the vote period has elapsed, the initiator of the
+vote, or their designee, closes it by replying to the thread with the vote
+results. That result email should use the same subject line preceded by
+`[RESULT][VOTE]`. Voting is carried out by replying to the vote mail and
+continues until the vote is closed. If a vote thread becomes inactive and
+remains open for too long, without a response from the initiator, the PMC Chair
```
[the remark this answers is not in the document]
ctubbsii: No. I intentionally left it open ended, based on expectations about the role of the PMC chair and to avoid being too legalistic in our approach to novel circumstances.

From https://www.apache.org/dev/pmc.html#chair :
> Remember that, as in any committee, the chair is a facilitator and their role within the PMC is to ensure that everyone has a chance to be heard and to enable meetings and mailing lists to flow smoothly.

It seems to follow that this is one way the activity on the mailing list can get held up, and the chair may have to step in to facilitate things flowing smoothly again. If we impose too strict rules around how long is too long, or under what circumstances, etc., then we're constraining the ability of the chair to facilitate things to keep things flowing smoothly. I've heard similar sentiments expressed about the chair's role from a variety of people, including advice from a board member once (years ago) that the PMC chair probably should have stepped in and made a decision when there was a stalemate conflict between contributors, even though there was nothing technically written that conflicts like that should be resolved in that way.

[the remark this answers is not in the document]
brianloss: Sure, that makes sense too. I was thinking about it from the other standpoint--would a chair be afraid to intervene because they're not sure when it's been long enough? However, I definitely agree that we don't want to be too legalistic, so I'm ok with this.

contributor/bylaws.md
```diff
@@ -5,146 +5,132 @@ redirect_from:
...
 
-Accumulo is a project of the [Apache Software Foundation][foundation]. The
-foundation holds the copyright on Apache code including the code in the
-Accumulo codebase. The [foundation FAQ][foundation-faq] explains the operation
-and background of the foundation.
-
-Accumulo is typical of Apache projects in that it operates under a set of
-principles, known collectively as the Apache Way. If you are new to Apache
+Accumulo is a typical ASF project, operating under a set of principles known
+collectively as [the Apache Way][thisistheway]. If you are new to ASF
```
DomGarguilo, line 16:
```suggestion
collectively as [The Apache Way][thisistheway]. If you are new to ASF
```

contributor/bylaws.md
```diff
@@ -154,162 +140,87 @@ An Accumulo release manager is also expected to:
...
-All dates in a plan are estimates, as unforeseen issues may require delays. The
-release manager may adjust dates as needed. In serious circumstances, the
-release manager may opt to call a re-plan vote.
+Accumulo follows a commit-then-review (CTR) policy. This means that consensus
+is not required prior to committing. Committers can make changes to the
+codebase without seeking approval beforehand, and those changes are assumed to
+be approved unless an objection is raised afterwards. Only if an objection is
+raised must a vote take place on the code change.
+
+However, just because committers can do this, it does not mean it is a always a
```
DomGarguilo, line 199:
```suggestion
However, just because committers can do this, it does not mean it is always a
```

contributor/bylaws.md
```diff
@@ -154,162 +140,87 @@ An Accumulo release manager is also expected to:
...
+good idea. Committers are expected to use their privileges responsibly and to
+minimize risks. Therefore, it is often is a good idea for committers to seek
```
DomGarguilo, line 201:
```suggestion
minimize risks. Therefore, it is often a good idea for committers to seek
```

contributor/bylaws.md
```diff
@@ -5,146 +5,132 @@ redirect_from:
...
+Because the PMC is a role within the Foundation, changes to PMC membership
+follow criteria established by the board. In general, this means that the board
+chooses who gets added or removed from the PMC, but they may delegate that
+authority to existing PMC members or the PMC chair. However, it is this
+project's practice to collectively make decisions about PMC additions among the
+existing PMC members on the project's private mailing list. Accordingly, the
+existing PMC members will conduct a vote before inviting new PMC members, so
+long as that vote can be done without violating the board's current procedures.
+Depending on the current board procedures, this vote may be merely advisory.
+Additionally, no vote will be conducted without first being preceeded by a
```
DomGarguilo, line 92:
```suggestion
Additionally, no vote will be conducted without first being preceded by a
```

DomGarguilo (changes requested):

contributor/bylaws.md
```diff
@@ -154,162 +140,87 @@ An Accumulo release manager is also expected to:
...
-decisions, only the votes of active committers are binding. Non-binding votes
-are still useful for those with binding votes to understand the perception of
-an action across the wider Accumulo community. For PMC decisions, only the
-votes of active PMC members are binding.
-
-See the [voting page][voting] for more details on the mechanics of voting.
+starting with `[VOTE]`. After the vote period has elapsed, the initiator of the
+vote, or their designee, closes it by replying to the thread with the vote
+results. That result email should use the same subject line preceded by
+`[RESULT][VOTE]`. Voting is carried out by replying to the vote mail and
```
DomGarguilo, line 166: > closes it by replying to the thread with the vote
results

> That result email should use the same subject line preceded by
`[RESULT][VOTE]`

Do these two conflict with each other? Is one suggesting replying to an email thread and another suggesting a new email thread with a new subject line?

dlmarion (approved):
Changes, including updates in comments, look good to me.

contributor/bylaws.md
ctubbsii: No, they aren't in conflict. A thread is not defined by the subject alone, and changing the subject is not the same as starting a new thread in the conventional sense or in the technical sense.

In the technical sense, you can reply to a thread and change its subject line before sending in most email clients. A good email client will preserve the threading headers so that the mailing list service will keep them threaded in spite of the subject change. Although they have evolved, these features have generally been around for decades and were enormously useful in the days when NNTP newsgroups were supported by popular email clients. Some email clients (from my observations, Outlook is one) are particularly bad at this and drop these threading headers, and effectively start a new thread every time you change the subject.

However, even if they aren't threaded in the technical sense, (such as if a user uses a bad email client, for instance) it's still the same thread in the conventional sense. Anybody looking at the subject line will be able to understand what's going on, because even if they are not technically identified as the same "thread" of conversation by the software, humans recognize the convention of adding tags to prefix a subject line as being part of the same overall thread of conversation.

An example post is [this one](https://lists.apache.org/thread/22fol574ch2dbp48w4wbb0hn31j2mx0m) whose [raw content](https://lists.apache.org/api/source.lua?id=22fol574ch2dbp48w4wbb0hn31j2mx0m) contains the header `In-Reply-To: <CAL5zq9bTLW8n6pJBuJpouwaycDeutEqE6C2sKTzLuTOFXDq2bw@mail.gmail.com>` which references [this previous message](https://lists.apache.org/thread/01lq7fs8pf3fy4tddsmoml9r9lgd9ogz)'s [header](https://lists.apache.org/api/source.lua?id=01lq7fs8pf3fy4tddsmoml9r9lgd9ogz) `Message-ID: <CAL5zq9bTLW8n6pJBuJpouwaycDeutEqE6C2sKTzLuTOFXDq2bw@mail.gmail.com>` as well as other messages in the same thread in the `References:` header.

DomGarguilo: Gotcha. Thanks for clearing that up. Makes sense now :+1:

DomGarguilo (approved):

ctubbsii:
I'm marking this ready for review to initiate a vote among the PMC to accept the state of this PR as of commit 3a7969deee281f13f3887a3c80da79727db2c326. No changes will be made to this PR while this vote is in progress.

PR #432, opened 2024-07-25
Add Kevin Rathbun PMC list

kevinrr888:
Adds Kevin Rathbun to PMC/committer list


keith-turner (approved):

ctubbsii:
@kevinrr888 For future reference, for the vast majority of merges of PRs for accumulo, you can actually just use squash merge (available from the dropdown on the merge button). That keeps the git history nice and linear and easier to follow later. For complicated things we might want to preserve more of the history, but for most discrete changes, squash is fine.

kevinrr888:
@ctubbsii Thank you for this reply. I thought this would be a good PR to merge in myself for the first time, and noticed afterwards that it was merged as two commits. Glad I'm learning this now. I will keep this in mind for future merges

PR #442, opened 2024-09-11
Add Metrics doc to User Manual configuration tab

DomGarguilo:
Adds the metric doc that is generated via the changes added in [accumulo PR #4850]
Screenshot from 2024-09-12 14-59-05


DomGarguilo:
This PR should probably wait to be merged until 3.1 is released as that is the branch that this metrics doc is generated off of.

dlmarion:
FYI that I did this for the 4.x docs already.

PR #451, opened 2025-03-04
Adds the lunr script to the local resources

ddanielr:
ASF started blocking external resources as of March 1st. This policy breaks the ability to pull the lunr script from the external hosting location.


ddanielr:
After applying this change, the only external resources I see being used are `www.apache.org` and `www.apachecon.com`

PR #452, opened 2025-03-22
Add updates for merge changes in Accumulo 4.0

cshannon:
Adds some initial information for users for the updated merge changes in Accumulo 4.0. I didn't want to go to detailed into the internals but the changes describe how the new merge is different (faster and fences files, etc compared to chop compactions) and also goes into auto merge and how to configure tablets using the API. I'm marking it as a draft as I'm sure some changes are needed and polishing but want to get some feedback on how things look and if we need more detail or not. I also assume there will be updates in the future before release as things are tweaked as 4.0 still is not close to release.


_docs-2/administration/merging.md
```diff
@@ -0,0 +1,89 @@
...
+2. The combined files must be less than `table.merge.file.max`
+3. The total size must be less than `table.mergeability.threshold`. This is defined as the combined size of RFiles as a percentage of the split threshold
+
+## Configuration
+
+The following properties are used to configure merging:.
+
+* `manager.tablet.mergeability.interval` -Time to wait between scanning tables to identify ranges of tablets that can be auto-merged (default is `24h`)
+* `table.mergeability.threshold` - A range of tablets are eligible for automatic merging until the combined size of RFiles reaches this percentage of the split threshold. (default is `.25`)
+* `table.merge.file.max` - The maximum number of files that a merge operation will process (default is `10000`)
```
keith-turner, line 34: This property applies to both api initiated merges and automatically initiated merges, the other two props only apply to automatic merges. Not sure if this is worth working in somehow, it was just something I noticed.

_docs-2/administration/merging.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Tablet Mergeability
```
keith-turner, line 36: May want to work in to this section what will happen on upgrade for existing tablets.

PR #453, opened 2025-04-04
Update tracing blog post

DomGarguilo:
This PR adds notes regarding updated setup for tracing. This is being added to the end of https://accumulo.apache.org/blog/2022/06/22/2.1.0-metrics-and-tracing.html

Here is what these additions look like:
image

The original notes that inspired this PR were created by @keith-turner


_posts/blog/2022-06-22-2.1.0-metrics-and-tracing.md
```diff
@@ -149,3 +149,24 @@ docker run -d --rm --name jaeger \
 8. View traces in Jaeger UI at http://localhost:16686. You can select the service name on the left panel and click `Find Traces` to view the trace information. If everything is working correctly, then you should see something like the image below.
 
 ![Jaeger Screenshot](/images/blog/202206_metrics_and_tracing/Jaeger_Screenshot.png)
+
```
dlmarion, line 152: The blog post was for the 2.1.0 release. It looks like the addition here updates / replaces items 5 and 6 above. I'm thinking that we should put a note in items 5 and 6 that says something like:

```
This contains instructions for configuring OpenTelemetry and Jaeger with Accumulo 2.1.0. As of Accumulo 2.1.4 the version of OpenTelemetry has been updated and the instructions have changed. See below for more information.
```

DomGarguilo: Good idea. Added in 4b88f58.

DomGarguilo: Screenshot from 2025-04-08 14-52-32
Screenshot from 2025-04-08 14-52-18

_posts/blog/2022-06-22-2.1.0-metrics-and-tracing.md
```diff
@@ -149,3 +151,27 @@ docker run -d --rm --name jaeger \
...
+## Note for Recent Accumulo Versions (April 2025)
+
+**Note:** This section replaces steps 5 and 6 in the [Tracing Example](#tracing-example) above. Follow steps 1-4 from the original instructions, then use the updated configuration below instead of steps 5-6, and continue with steps 7-8 to view traces in Jaeger.
```
ctubbsii, line 157: This all seems fine, but is this really a note for recent Accumulo versions, or is it a note for newer OpenTelemetry versions? Somebody could use the newer OpenTelemetry libraries with the older Accumulo releases, in which case, these instructions would apply, right? Likewise, somebody could use a newer version of Accumulo but with an older version of OpenTelemetry, in which case the older instructions would apply?

I think the instructions are clear, but it is not clear to me that the conditions under which you use them are correctly or clearly being communicated. I think the version of Open Telemetry is much more important than the version of Accumulo for this. And, I think it's wrong to assume that the versions in Accumulo's POMs are the authoritative source for what the user is (or should be) running.

Users are responsible for their own CLASSPATH, including updating them, even for older versions of Accumulo, in response to CVEs, bug fixes, dependency convergence, integration issues, etc. That is a downstream responsibility. So, we should be careful to word things without making assumptions that the dependencies we reference in our POM are what users are, or should be, using.

It might be more correct and clear to say something like: "Accumulo X was released compiled and tested with OpenTelemetry Y, which uses the Z property to configure...", rather than say "For Accumulo X, you need to set Z property ...."; the latter might be true, but only for very specific assumptions.

ctubbsii (commented):
An alternative to having an addendum, would be just to edit the page replacing the old instructions with the current ones, and have a note that says something short that says:

```
These instructions were updated from an earlier version to work with Accumulo 2.1.4 and OpenTelemetry 1.48.
```

Or, instead of modifying the document in place, you can release a copy of the blog post with a newer date, with those same edits, and link back to this one as the earlier version. That way, the earlier version is left unmodified, in case anybody wants to easily reference it, and we get more succinct current instructions that aren't bloated with the older ones as well.

DomGarguilo:
@ctubbsii I tried to address your comments in e4ec9cd.

> An alternative to having an addendum, would be just to edit the page replacing the old instructions with the current ones, and have a note that says something short that says:
> 
> ```
> These instructions were updated from an earlier version to work with Accumulo 2.1.4 and OpenTelemetry 1.48.
> ```
> 
> Or, instead of modifying the document in place, you can release a copy of the blog post with a newer date, with those same edits, and link back to this one as the earlier version. That way, the earlier version is left unmodified, in case anybody wants to easily reference it, and we get more succinct current instructions that aren't bloated with the older ones as well.

I think I prefer the addendum approach since it preserves the old instructions which some may still find helpful while also adding the new stuff without causing too much churn.

_posts/blog/2022-06-22-2.1.0-metrics-and-tracing.md
```diff
@@ -85,7 +85,9 @@ general.micrometer.factory=org.apache.accumulo.test.metrics.TestStatsDRegistryFa
 
 10. Log into Grafana (http://localhost:3003/) using the default credentials (root/root). Click the `Home` icon at the top, then click the `Accumulo Micrometer Test Dashboard`. If everything is working correctly, then you should see something like the image below.
 
-![Grafana Screenshot](/images/blog/202206_metrics_and_tracing/Grafana_Screenshot.png)
+<a class="p-3 border rounded d-block" href="{{ site.baseurl }}/images/blog/202206_metrics_and_tracing/Grafana_Screenshot.png">
+   <img src="{{ site.baseurl }}/images/blog/202206_metrics_and_tracing/Grafana_Screenshot.png" class="img-fluid rounded" alt="Grafana Screenshot"/>
+</a>
```
DomGarguilo, line 90: While I was making changes to this file I decided to add the styling we use in other pages for these screenshots. This helps the images remain in line with the other content.

For example:
image

dlmarion (approved):

_posts/blog/2022-06-22-2.1.0-metrics-and-tracing.md
```diff
@@ -148,4 +152,30 @@ docker run -d --rm --name jaeger \
...
+<a class="p-3 border rounded d-block" href="{{ site.baseurl }}/images/blog/202206_metrics_and_tracing/Jaeger_Screenshot.png">
+   <img src="{{ site.baseurl }}/images/blog/202206_metrics_and_tracing/Jaeger_Screenshot.png" class="img-fluid rounded" alt="Jaeger Screenshot"/>
+</a>
+
+## Updated Configuration with Newer Versions of OpenTelemetry
+
+The original tracing example above was written for Accumulo 2.1.0 which uses OpenTelemetry 1.19.0. Accumulo 2.1.4 has been updated to use OpenTelemetry 1.48.0, which requires different configuration. When working with Accumulo 2.1.4 or any deployment using OpenTelemetry 1.48.0 or newer, use OTLP (OpenTelemetry Protocol) instead of the direct Jaeger exporter.
+
+Follow steps 1-4 from the [Tracing Example](#tracing-example) section, then use these updated steps in place of steps 5-6:
+
```
ctubbsii, line 164: Kinda late to mention this, if the user was going step-by-step. Should probably mention which specific steps to skip in the initial Note above, before they get this far. Otherwise, the first time you mention to skip steps 5-6 occur after the user has possibly already done them.

_posts/blog/2022-06-22-2.1.0-metrics-and-tracing.md
ctubbsii: It doesn't matter, really, but I think there's probably a way to specify the class on the image or at least on the section (div or p) containing the image, in the markdown form, that might be easier to read or maintain. Markdown, in general, is easier to work with than the raw HTML, but like I said, this doesn't really matter. The raw HTML is fine.

PR #460, opened 2025-04-30
Improve development dockerfile and associated readme

DomGarguilo:
* breaks some commands into individual lines to make future maintenance less disruptive
* use uppercase `AS` in dockerfile. There was a warning regarding inconsistent case
* improve readme with a pitfall that I ran into
* remove duplicate line from readme and empty lines

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 
 FROM ruby:3.2.2-slim-bullseye AS base
 
-RUN apt-get update && apt-get install -y --no-install-recommends \
+RUN apt update && apt install -y --no-install-recommends \
     build-essential \
     git \
     curl \
@@ -26,8 +26,8 @@ RUN gem update --system \
   && bundle install \
   && gem cleanup
 
-ENV HOST=0.0.0.0 \
-    PORT=4000
+ENV HOST=0.0.0.0
+ENV PORT=4000
 
 EXPOSE $PORT
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # making updates to the accumulo website without requiring the dev
 # to maintain a local ruby development environment.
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -17,16 +17,17 @@ WORKDIR /mnt/workdir
...
-COPY Gemfile.lock ./Gemfile.lock
+COPY Gemfile Gemfile.lock ./
 
 # Gems will be installed under GEM_HOME which is set by the ruby image.
 # See https://hub.docker.com/_/ruby for details.
 
-RUN gem update --system && bundle install && gem cleanup
+RUN gem update --system \
+  && bundle install \
+  && gem cleanup
```
ctubbsii, line 27: I don't know how Dockerfiles are parsed, but in bash, the backslash is redundant if the line ends with `&&`, as in:

```
RUN gem update --system &&
  bundle install &&
  gem cleanup
```

This is even easier to read, but I don't know if it works the same in Dockerfiles.

Dockerfile
```diff
@@ -17,16 +17,17 @@ WORKDIR /mnt/workdir
...
 
-ENV HOST=0.0.0.0
-ENV PORT=4000
+ENV HOST=0.0.0.0 \
+    PORT=4000
```
ctubbsii, line 30: I think it's better as two separate `ENV` lines. They are two independent environment declarations. This new syntax makes it seem like the two are related in some way, and might be confused with something like `ENV PROPWITHEMBEDDEDSPACE=0.0.0.0\ 4000`

Dockerfile
DomGarguilo: I tried this out and it did not work so I guess docker requires it.

Dockerfile
DomGarguilo: Reverted in 2a11117

PR #461, opened 2025-05-01
Started draft 4.0.0 release notes. Added table of API changes.

dlmarion:


dlmarion:
https://github.com/apache/accumulo/pull/5527 details how I generated the APi changes. I figure these can be useful to determine what needs to be marked as deprecated for a 3.1 release and to help jog our memory on changes for the final version. The generated table doesn't look that great, so maybe we don't keep it in the final version.

DomGarguilo:
I spun these changes up and things look good. The new API changes table is a bit confusing though.

image
A lot of the entries aren't really meaningful or are lacking details about what they are referring too. Also the `SEVERITY` and `OLD` column headers are not very clear to me.

Also since the table is so large the sections below it are sort of hidden. The table might be better off at the bottom as to not bury any other sections.

dlmarion:
@DomGarguilo  - Regarding the table,I agree it's not great. Maybe some css changes would make it look better? However, maybe it's not an issue. Talking with Christopher, he thinks that we should not put this in the release notes. I think capturing the differences will be useful though for our own internal use. It will be useful when working on the 3.1.0 release. I'm thinking that maybe I just remove the link in the release notes to the API changes file, and move the api changes file somewhere else in the directory structure.

rashworld-max (approved):

PR #469, opened 2025-08-18
Update tracing blog post to remove the instructions to set the client property

### LINKED ISSUE #465: Update tracing blog post to fix client side tracing properties
The blog post tells the reader to enable tracing via the client properties however this does not actually do anything so should be removed/updated.

This change was brought up here: https://github.com/apache/accumulo/pull/5719#issuecomment-3059999066

ibilley7:
Removed step 4 under the Tracing Example as setting that property does nothing anymore.

Closes #465


DomGarguilo (approved):
There was something going on where the numbered list was restarting at 1 

Screenshot from 2025-08-19 15-18-26

I pushed a commit to fix this and now everything LGTM

PR #488, opened 2026-09-08
Created Scan Server blog post

### LINKED ISSUE #4463: Write up scan server example for Blog post
**Is your feature request related to a problem? Please describe.**
We currently don't have great documentation for using scan servers. 

**Describe the solution you'd like**
A blog post should be written up and published for using scan servers and the ScanSelector. 

https://github.com/apache/accumulo/blob/2.1/server/tserver/src/main/java/org/apache/accumulo/tserver/ScanServer.java#L130 

https://github.com/apache/accumulo/blob/main/core/src/main/java/org/apache/accumulo/core/spi/scan/ConfigurableScanServerSelector.java#L150

dlmarion:
Closes https://github.com/apache/accumulo/issues/4463

Assisted-by: OpenAI GPT-5.6 Sol

### DIFF
--- a/_posts/blog/2026-09-08-scan-servers.md
+++ b/_posts/blog/2026-09-08-scan-servers.md
@@ -1,0 +1,300 @@
+---
+title: "Accumulo Scan Servers"
+author: Dave Marion
+---
+
+## Introduction
+
+Apache Accumulo Tablet Servers perform many tablet maintenance functions (read, write, compacting, merging, splitting, etc.). Since Tablet Servers typically host many tablets these tablet maintenance functions can be queued waiting for resources or interrupted when the Manager decides to migrate a tablet or if the Tablet Server process dies. Scan Servers, like Compactors, move processing out of the Tablet Server to help mitigate some of these situations. A Scan Server is an optional, read-only Accumulo process that handles eventually consistent scans independently of Tablet Servers. Administrators can deploy and scale Scan Servers according to scan demand, providing additional read capacity without adding the responsibilities of hosting tablets or processing writes. Beginning in Accumulo 2.1.5, Scan Servers can also scan offline tables without bringing those tables online.
+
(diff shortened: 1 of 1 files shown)


_posts/blog/2026-09-08-scan-servers.md
```diff
@@ -0,0 +1,275 @@
...
+      {"servers": "3", "busyTimeout": "5s"},
+      {"servers": "100%", "busyTimeout": "30s"}
+    ]
+  }
+]
+```
+
+### Controlling Latency
+
+In 4.0.0 the property `table.compaction.minor.age` replaces `table.compaction.minor.idle`. The latter property only triggered a minor compaction when writes to the Tablet had stopped and the tablet was idle. The `minor.age` property now considers the oldest in-memory mutation for the tablet and triggers a minor compaction even with writes continuing.
```
phrocker: is there a default to mention here?

phrocker (approved):
Nice writeup — the upgrade checklist in particular is the thing I'd have wanted when moving a 2.1 deployment forward. I checked the property tables against rel/2.1.0 and current main and they line up to the best of my ability. 

Two things I'd call out before merge: the offline-table section says to set availability to NEVER, but the TabletAvailability values are HOSTED/ONDEMAND/UNHOSTED — NEVER is a TabletMergeability value, so I think UNHOSTED is what's meant?

I think that's a nit so I'm approving otherwise.

_posts/blog/2026-09-08-scan-servers.md
dlmarion: Yeah, I can mention that and fix the NEVER -> UNHOSTED.

DomGarguilo (approved):

_posts/blog/2026-09-08-scan-servers.md
```diff
@@ -0,0 +1,275 @@
...
+
+Using Scan Servers comes with an explicit consistency tradeoff. Tablet Servers can serve immediate scans using both persisted files and data still held in memory. Scan Servers instead read persisted RFiles, so an eventual scan may not include recent writes that have not yet been flushed. Cached tablet metadata can also affect when newly created files become visible. Scan Servers are therefore a strong fit for reporting, historical analysis, and background processing, but not for operations that require read-after-write consistency.
```
ddanielr: I think it would be important to point out the drawback of the tablet servers being restricted to only performing work for their hosted tablets. 

There's a consistency tradeoff, but a huge upside for parallel read actions on the same tablet across multiple Scan Servers.

PR #490, opened 2026-09-10
Add Gen-AI instructions to Coding Guidelines

dlmarion:


dlmarion:
I looked at some commit message in Apache repositories to see how other projects are handling gen-ai attribution in commit messages. Here is one from [Thrift](https://github.com/apache/thrift/commit/1b7e0dfd83112be81999d4eb27d4b67c76c9c9f8), [Hadoop](https://github.com/apache/hadoop/commit/40433c51b4826127535c76035a970ba688cac82b) and [Airflow](https://github.com/apache/airflow/commit/2bb32e370cce86b1c422a249b95af1a0f743517b).

My suggestion to **not** use `Co-Authored-By` as that format [should](https://docs.github.com/en/pull-requests/how-tos/commit-changes/creating-a-commit-with-multiple-authors) include an email address and maybe we save that attribution for humans.

ddanielr (approved):
I agree with the `Assisted-By` or `Generated-by` commit message format and leaving `Co-Authored-By` for human authors.  

If we wanted to standardize on a single one then I would vote for using `Assisted-By`.

phrocker:
I think https://www.apache.org/legal/generative-tooling.html#include-in-contributions suggests Generated-by: <tool> <version> for provenance. 

Assisted-by: leans toward claiming more human authorship while disclosing less provenance. Generated-by: makes no authorship claim at all, it just names a tool, which is the cheaper and more honest thing to assert.

dlmarion:
> I think https://www.apache.org/legal/generative-tooling.html#include-in-contributions suggests Generated-by: for provenance.
> 
> Assisted-by: leans toward claiming more human authorship while disclosing less provenance. Generated-by: makes no authorship claim at all, it just names a tool, which is the cheaper and more honest thing to assert.

I suggested two options because I think there is a distinction. To me `Generated-By` implies that the machine did all the work. This could be true in some cases where the developer uses an AI harness-only to iterate and modify code and the human just does review. But, in the case where the developer only uses a machine to do part of the work `Generated-By` implies that the developer did little to no work.

Maybe I'm taking it too personally :-).

I re-read the paragraph that uses `Generated-By` as an example and if the requirement is for provenance and traceability rather than some type of assignment of authorship, then maybe a more generic tag can be used, say `AI-Model-Used:`.

DomGarguilo:
> To me `Generated-By` implies that the machine did all the work. This could be true in some cases where the developer uses an AI harness-only to iterate and modify code and the human just does review. But, in the case where the developer only uses a machine to do part of the work `Generated-By` implies that the developer did little to no work.

This is how `Generated-By` vs `Assisted-By` reads in my head too. In terms of what is most correct/best practice here I am not sure.

> If we wanted to standardize on a single one then I would vote for using Assisted-By.

This suggestion sounds reasonable to me too

ddanielr:
> I suggested two options because I think there is a distinction. To me `Generated-By` implies that the machine did all the work. This could be true in some cases where the developer uses an AI harness-only to iterate and modify code and the human just does review. But, in the case where the developer only uses a machine to do part of the work `Generated-By` implies that the developer did little to no work.

I was also thinking that `Generated-by` has connotations of minimal human involvement.

> I think https://www.apache.org/legal/generative-tooling.html#include-in-contributions suggests Generated-by: for provenance.
> 
> Assisted-by: leans toward claiming more human authorship while disclosing less provenance. Generated-by: makes no authorship claim at all, it just names a tool, which is the cheaper and more honest thing to assert.

I see your point about `Assisted-by`. I'm mostly thinking of this wording in the context of PR reviews. 
In a perfect environment all reviews are getting the same level of testing and review but that's generally not the case.

Personally, It would be helpful to know which files/commits were completely generated code so I know what areas of the code review to focus on.

As an example, we have thrift generate our rpc classes and generally that code is not reviewed. Instead we focus on the behavior tests passing. 

The only specific check that does occur ensures the generated content has not changed unintentionally.

I accept that I might be thinking about this too much. 
If we don't want to be that granular then maybe `AI-Model-Used:` is a good way of bridging that gap?

phrocker:
> > I suggested two options because I think there is a distinction. To me `Generated-By` implies that the machine did all the work. This could be true in some cases where the developer uses an AI harness-only to iterate and modify code and the human just does review. But, in the case where the developer only uses a machine to do part of the work `Generated-By` implies that the developer did little to no work.
> 
> I was also thinking that `Generated-by` has connotations of minimal human involvement.
> 
> > I think https://www.apache.org/legal/generative-tooling.html#include-in-contributions suggests Generated-by: for provenance.
> > Assisted-by: leans toward claiming more human authorship while disclosing less provenance. Generated-by: makes no authorship claim at all, it just names a tool, which is the cheaper and more honest thing to assert.
> 
> I see your point about `Assisted-by`. I'm mostly thinking of this wording in the context of PR reviews. In a perfect environment all reviews are getting the same level of testing and review but that's generally not the case.
> 
> Personally, It would be helpful to know which files/commits were completely generated code so I know what areas of the code review to focus on.
> 
> As an example, we have thrift generate our rpc classes and generally that code is not reviewed. Instead we focus on the behavior tests passing.
> 
> The only specific check that does occur ensures the generated content has not changed unintentionally.
> 
> I accept that I might be thinking about this too much. If we don't want to be that granular then maybe `AI-Model-Used:` is a good way of bridging that gap?

I wouldn't say I would have very much of a strong reaction to Assisted-By or Generated-By, but AI-Model-Used avoids any reference to how the code was generated and it's quite binary about its intent, which I like. It's a good discussion point to see if anyone else disagrees, but I like AI-Model-Used for its simplicity.

ddanielr (approved):

pages/how-to-contribute.md
```diff
@@ -60,6 +60,9 @@ Any questions/ideas don't hesitate to [contact us][contact].
 * Accumulo follows [semver] for its [public API](/api/).
 * Every file requires the ASF license header as described in [ASF Source Header][srcheaders].
 * Do not use Author Tags. The code is developed and owned by the community.
+* If you used Generative-AI to generate or assist in developing the change, then please be sure to:
+    * Follow the guidelines at https://www.apache.org/legal/generative-tooling.html
+    * Add a line near the end of your commit message containing `AI-Model-Used: <model>`.
```
ctubbsii, line 65: The ASF guidelines are still evolving, and are currently in draft. While the legal page is binding, and shows an example, there are additional draft best practices at https://rai.apache.org/best-practices.html

Much of this is currently redundant, and conflicting. I have read it over and provided feedback to the RAI committee. My personal recommendation is that we should use `Co-Authored-By: <model>`. While the standard format of the Author and Committer fields include an email address, it is actually optional and not strictly required by git. The "Co-Authored-By" tag inside the body of the commit follows the same format, so it is also optional to include an email address, and I would just omit them, since there isn't an email for an AI model.

The RAI best practices draft pages also document other alternatives, which I think are less good. I don't believe "AI-Model-Used" is one of them, and I would strongly advise that we not invent something new. I think sticking to "Co-Authored-By" is the most obvious and already understood by existing tooling. It also covers many other cases (assisted-by, generated-by, generated-then-human-reviewed, reviewed-by, etc.), because nobody expects this tag to communicate the specific quantity and nature of each of the contributions of each author, and nobody is asking for that because that'd be overkill. Let's keep it simple, and jut use `Co-Authored-By: <model>`

ctubbsii (changes requested):
-1 to these changes. (See my earlier comment)

pages/how-to-contribute.md
dlmarion: I have no issue changing `AI-Model-Used` to something else. Do we have an idea of when the ASF guidelines will come out of draft? Looking at https://rai.apache.org/authored-by-tags.html, the draft is recommending `ai-assisted-by` and possibly `Apache-ai` tags.

ctubbsii: No idea when it will be final. But in discussion with one of the RAI members, they said they were going to update them soon, and that I would probably be happier with the changes (after I provided feedback similar to here). Even after they are final, it will still just be recommendations, not requirements. So, we can do whatever we want. The most important thing is just marking that it was used by some mechanism. The "how" doesn't matter except for personal/group preference. For our contributor guidelines, I'd keep things very minimal... just a link to ASF legal policy, a possible link to the RAI draft guidelines (just the top page, since the pages might change) with a warning that they are subject to change, and maybe a general suggestion to use "Co-Authored-By: or a similar acknowledgment in the commit message, with information about the AI model used", rather than dictate a specific format.

pages/how-to-contribute.md
```diff
@@ -60,6 +60,10 @@ Any questions/ideas don't hesitate to [contact us][contact].
 * Accumulo follows [semver] for its [public API](/api/).
 * Every file requires the ASF license header as described in [ASF Source Header][srcheaders].
 * Do not use Author Tags. The code is developed and owned by the community.
+* If you used Generative-AI to generate or assist in developing the change, then please be sure to:
+    * Follow the legal guidelines at https://www.apache.org/legal/generative-tooling.html
+    * Consult the (draft) guidelines or best practices at: https://rai.apache.org/
+    * Add the AI model name and version used in your commit message, preferably in a tag
```
ctubbsii:
```suggestion
    * Add the AI model name and version used in your commit message, preferably in a well-known tag (e.g. "Co-Authored-By:", or less commonly "Assisted-By:", or similar)
```

ctubbsii (commented):
It's fine as-is. Could make this one change, but it's fine either way:

ctubbsii (approved):

ctubbsii:
FWIW, "AI-Assisted-By" is growing on me (it's what gpt 5.6 advises when asked). The more common growing consensus from the subset of the internet I've read on this subject lately is a simpler "Assisted-by" trailer, but some sites also use this for humans as well. The "assisted-by" variants are growing on me (esp. "AI-Assisted-By"), but I still don't see much difference between an "assist" with authoring something and a "co-author", not enough to justify a new convention anyway.

However, I do wish there was a standard identification format for models, like SPDX exists for licenses. Right now, every provider seems to use a different naming format, so it's not so easy to tell if "GPT 5.6 (luna)" is the same as "OpenAI 5 (2025, small)" [NOTE: this is a contrived example; I made these version strings up].

phrocker:
> FWIW, "AI-Assisted-By" is growing on me (it's what gpt 5.6 advises when asked). The more common growing consensus from the subset of the internet I've read on this subject lately is a simpler "Assisted-by" trailer, but some sites also use this for humans as well. The "assisted-by" variants are growing on me (esp. "AI-Assisted-By"), but I still don't see much difference between an "assist" with authoring something and a "co-author", not enough to justify a new convention anyway.
> 
> However, I do wish there was a standard identification format for models, like SPDX exists for licenses. Right now, every provider seems to use a different naming format, so it's not so easy to tell if "GPT 5.6 (luna)" is the same as "OpenAI 5 (2025, small)" [NOTE: this is a contrived example; I made these version strings up].

Yah this is still early times and I suspect it will change in time. The lack of convergence is a sign to monitor until some standard appears if it ever does.

ctubbsii:
Apparently [SPDX 3 is going with purl](https://spdx.github.io/spdx-spec/v3.0.1/annexes/pkg-url-specification/), and has a full JSON spec for an AI model profile. The purl one-liner might look something like `pkg:aws/bedrock/us/openai/gpt-5.6-luna`, for example. I'm not sure if this will catch on for git commits, though.

PR #491, opened 2026-09-11
Updated Accumulo 4.x Configuration docs

dlmarion:


_docs-4/configuration/overview.md
```diff
@@ -72,6 +74,34 @@ client.instanceOperations().modifyProperties(properties -> {
...
+indicated in the **zk mutable** section of their description. Resource Group configuration can be set using the following shell command:
+
+```console
+config -rg GROUP_NAME -s PROPERTY=VALUE
+```
+
+They can also be set using {% jlink org.apache.accumulo.core.client.admin.ResourceGroupOperations %} in the Java API:
+
+```java
+client.resourceGroupOperations().setProperty("table.durability", "flush");
```
ddanielr, line 91: This example is using a table property instead of a server property. 
I think this is actually a case where we would want a failure.

_docs-4/configuration/overview.md
```diff
@@ -72,6 +74,34 @@ client.instanceOperations().modifyProperties(properties -> {
...
+```
+
+The java api also supports adding, modifying and removing multiple properties in a single operation:
+
+```java
+client.resourceGroupOperations().modifyProperties(properties -> {
+  properties.remove("table.file.max");
+  properties.put("table.bloom.enabled", "true");
+  properties.put("table.bloom.error.rate", "0.75");
+  properties.put("table.bloom.size", "128000");
```
ddanielr, line 101: These are all table properties.

```suggestion
client.resourceGroupOperations().modifyProperties(properties -> {
  properties.remove("general.block.cache.manager.class");
  properties.put("general.low.mem.protection.scan", "true");
  properties.put("general.server.iter.opts.compression", "zstd");
  properties.put("tserver.summary.retrieval.threads", "128");
```

PR #492, opened 2026-09-16
Fix outdated base image in ContainerFile

Amemeda:
While setting up a container environment using [accumulo website README](https://github.com/apache/accumulo-website#testing-using-a-container-environment), found that `ContainerFile` is using outdated base image `FROM ruby:3.2.2-slim-bullseye AS base`, causing the following 2 podman commands to fail:
- `podman build -t webdev -f Containerfile .`
- `podman run -i -t --rm -v "$(pwd)":/mnt/workdir:Z -p 4000:4000 webdev`

The `bullseye` base image is an older debian version nearing end-of-life. And the run command is expecting the registry name, results in the following error;
`Trying to pull docker.io/library/webdev:latest...
Error: initializing source docker://webdev:latest: reading manifest latest in docker.io/library/webdev: requested access to the resource is denied`

Fixed this by changing the base image in `ContainerFile` to use `bookworm`, and adding the registry name to the base image: `FROM docker.io/library/ruby:3.2.2-slim-bookworm AS base`.


DomGarguilo (approved):
Tested and thing LGTM

